PR #24, opened 2010-12-30
File munge api

kentfredric:
This is mostly a request for a review of ideas and implementation. 

If you like the concept , or think it needs work, get back to me =). 

the idea is to unify the way people munge files, which currently has to be done differently depending on whether its a file from code, or a string file.

This doesn't appear to be a problem much at the moment, but if you try making a test or anything that gets moved/munged a File::FromCode, it will explode in a ball of fire when the munger tries to do 

$file->content( $foo ) and the FromCode file is immutable in that context.

The Munge API simplifies this by making consumers interact with the content via a callback fashion, forcing a sub-ref, so the underlying infrastructure can do the right thing, and either modify the content _now_, or use coderef chaining so the files content is aggregated when its needed.

```
$file->munge(sub{ 
    my( $fileself, $content ) = @_; 
    return munged($content);
});
```

On normal files, the code runs immediately, on FromCode files, it chains it so its more like this:

```
my $orig = $file->{coderef}; 
$file->{coderef} = sub { 
     my $content = $orig->(); 
     return munge( $content );
};
```

### DIFF
--- a/lib/Dist/Zilla/File/FromCode.pm
+++ b/lib/Dist/Zilla/File/FromCode.pm
@@ -3,7 +3,8 @@ package Dist::Zilla::File::FromCode;
 
 use Moose;
 use Moose::Util::TypeConstraints;
-
+use Params::Util qw( _CODELIKE );
+use Carp qw();
 use namespace::autoclean;
 
 =head1 DESCRIPTION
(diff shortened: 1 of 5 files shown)


kentfredric:
I have a new idea here that might be more useful: 

Instead of making the file nodes themselves contain mungers, the MungeFile role could have a stub like this, similar to its `->munge_file( $file ) for @files;`
```perl
sub munge_file_content {
      my ( $self, $file, $encoding, $content ) = @_;
      return unless $file->name =~ /expr/;   #  No munge.
      return {  # Some kind of munging
         encoding => $encoding,
         content   => $content,
      };
}
```

I'm not sure if that design is viable or not, it seems to me there needs to be some kind of "mark for munging before actually munging it" step there

rjbs:
This idea was never obviously terrible, but it never seemed viable to me, and I am now closing it.

PR #37, opened 2011-04-12
add a PreviousVersionProvider role

yanick:
Right now, plugins implementing the VersionProvider role
are doing up to two different tasks: retrieving the previous
version number, and incrementing it according to a given scheme.
The goal of the PreviousVersionProvider role is to split those
tasks between the two roles.

Of course, I'm not submitting the patch without having some secret
agenda: I seek this new role as a way to cleanly split the semantic
versioning scheme and the Git component that I'm using in
http://babyl.dyndns.org/techblog/entry/dist-zilla-semanticversion


rjbs:
Is this still relevant?

rjbs:
I am closing this as irrelevant/abandoned. Please re-open if this is not the case.

PR #55, opened 2011-09-02
Additional configuration for "follow_symlinks" 

SamuraiJack:
- enable additional configuration for "follow_symlinks" use case

  without "follow_skip" GatherDir may die when processing some
  file/dir for the second time

Please see http://search.cpan.org/~jesse/perl-5.14.1/lib/File/Find.pm#follow_skip for exact details.


rjbs:
I am closing this on the assumption that it is abandoned.

PR #68, opened 2011-11-03
Add diag_log method and auto_diag option

madsen:
I was thinking that something like diag_log ought to be in core.  I came up with a nicer way of doing it.  You just add `auto_diag => 1` to the tester parameters when creating your `$tzil` object.  That causes it to record the current number of failed tests.  When `$tzil` is destroyed, if there have been additional test failures, it writes out the log messages.

`$tzil->diag_log` passes the log entries to Test::Builder's diag()

`auto_diag => 1` does that automatically if any tests fail between construction and destruction

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,11 @@
 Revision history for {{$dist->name}}
 
 {{$NEXT}}
+          Dist::Zilla::Tester objects now have a diag_log method to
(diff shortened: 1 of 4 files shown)


rjbs:
I will apply this if rebased.

PR #69, opened 2011-11-07
Critic workaround, now with tests and an actual working implementation

DarwinAwardWinner:
Much better than [last time](https://github.com/rjbs/dist-zilla/pull/27).


rjbs:
closed in favor of #168

PR #73, opened 2011-12-07
Here's some code/test/doc for making NextRelease have a configurable delim

mateu:


rjbs:
I think that this still seems worth doing, with the changes I descibed.  Marked "needs rework."

PR #121, opened 2012-09-05
Add `build_element` support to the ModuleBuild plugin.

theory:
This is so that, if you have non-Perl files in `lib` (`.js`, `.sql`, `.jpg`, or whatever), you can tell Module::Build to install them.

I did a bit of testing with MakeMaker, and it looks as if it installs whatever is in `lib` without having to be told.


rjbs:
I apologize for this patch languishing.  I will apply it post haste if rebased.  (I had a quick go at it, but it was not trivial.)

karenetheridge:
I don't think it's a good idea for us to be supporting putting non-perl files into PERL5LIB. That's what a sharedir is for.

jberger:
Share dir does do this but at additional cognitive overhead.

theory:
Rebased.

karenetheridge:
> but at additional cognitive overhead.

@jberger what do you mean by this?

theory:
@karenetheridge This is a feature of Module::Build. You may not want to encourage it, which I get, but then that should go into the Module::Build docs as a recommendation or something (and the EU::MM docs). Ideally, Dist::Zilla would pass through whatever the installer modules support.

karenetheridge:
@leont what do you think? Should we try to discourage this feature of Module::Build, or live with it?

jberger:
ShareDir still has the problem of the files being in a different place when you are in a git checkout or after installation. To use it people have to dream up heuristics to decide which scenario they are in. Allowing to install non perl files directly is useful in that you don't need these heuristics.

karenetheridge:
> ShareDir still has the problem of the files being in a different place when you are in a git checkout or after installation.

Test::File::ShareDir works pretty well (although its guts need a lot of love.. different topic..)

Leont:
build_element doesn't map one-on-one with sharedir, really; it's more that its default action kind of does, but typically that's not a particularly interesting thing to do with it. Discouraging it is silly, encouraging sharedirs when they make sense is sensible, there two are not particularly related in my mind.

rjbs:
Thanks, applied!

karenetheridge:
Oops, there's no documentation for this new option!

PR #133, opened 2012-12-07
please remove copyright year from the distini plugin.

xenoterracide:
using the current year of release is more likely the desired behavior, and
will keep the latest release under copyright longer than old ones.

Signed-off-by: Caleb Cushing <xenoterracide@gmail.com>


rjbs:
This hasn't come up again, and while I may yet do what I suggested, I don't see it happening very soon.

PR #135, opened 2012-12-26
loosen the constraint on copyright_year

book:
given that the value is inserted in a string, and not used in any
computation involving integers, it seems nice to give authors the
option to declare a range of years is they wish to, such as

```
copyright_year = 2010-2012
```


rjbs:
Thanks, applied.
…
WITH A SOUR LOOK ON MY FACE

😊

rjbs:
12yr later, I take book's suggestion #745

PR #136, opened 2013-01-06
DZP/PkgVersion: v5.14+ package Name BLOCK; notation support.

kentfredric:
First pass working attempt at injecting a version declaration inside the BLOCK.

Note this offers no real support for << package Name v3.510 BLOCK >> form, and will just ignore that stated version
and put a (possibly differnet) version after it inside the BLOCK
```perl
package Foo v1.2 { }
```
-->
```perl
package Foo v1.2 {
{
	$Foo::VERSION = '1.3';
}
}
```
^^ Possible at present.

And I intentionally left support out for automatically sticking the version in the 

```perl
package Foo { 
```

Line automatically, as I feel thats the job for another plugin.

### DIFF
--- a/lib/Dist/Zilla/Plugin/PkgVersion.pm
+++ b/lib/Dist/Zilla/Plugin/PkgVersion.pm
@@ -72,6 +72,25 @@ has die_on_existing_version => (
   default => 0,
 );
 
+sub _make_version_declaration {
+  my ($self,$package) = @_;
+  my $version = $self->zilla->version;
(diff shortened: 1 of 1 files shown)


rjbs:
I like the idea of supporting the newer forms of `package`, but I'd like to get it all right at once.  If needed, I can try to produce a loose spec.

karenetheridge:
I suspect better support might be needed in PPI (for detection of the package BLOCK syntax)?

PR #138, opened 2013-01-19
Make dzil test more generic

shlomif:
These two commits allow to specify a different command to run instead of "test" in "make test" or "./Build test".


rjbs:
This can't be used from dist.ini, as LeonT noted.  You didn't make any case for why this would be useful, anyway. Closing pending being convinced and probably subsequent rework.

PR #146, opened 2013-02-10
Adding custom (displayed) target for CPAN::Uploader (and user/pass prompts)

Getty:
Trivial change that just adds an attribute target on UploadToCPAN that allows to set a custom text displayed instead of PAUSE. This is a followup on the already added change to CPAN::Uploader (target option)


rjbs:
* attribute should be called destination_name
* should default to PAUSE, rather than use a predicate, unless I'm confused?
* not sure I see why target is being passed to the Uploader

With those changes, would apply.

karenetheridge:
Getty tells me this was for DDG and is therefore no longer needed.

rjbs:
@Getty Shall I close this?

PR #164, opened 2013-04-07
MakeMaker's NAME must be a valid package name, not ditribution name replacing - with ::

miyagawa:
`$self->zilla->name` contains a distribution name and it might be different from "main_module" package names, which is required by MakeMaker's NAME.

From perldoc ExtUtils::MakeMaker:

    "NAME" must be a valid Perl package name and it must have an
    associated ".pm" file. For example, "Foo::Bar" is a valid "NAME" and
    there must exist Foo/Bar.pm.

cc @dagolden

### DIFF
--- a/lib/Dist/Zilla/Plugin/MakeMaker.pm
+++ b/lib/Dist/Zilla/Plugin/MakeMaker.pm
@@ -173,11 +173,22 @@ sub share_dir_code {
 sub main_module_name {
   my $self = shift;
 
-  my $meta = Module::Metadata->new_from_file($self->zilla->main_module->name);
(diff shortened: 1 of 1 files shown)


rjbs:
I apologize for letting this languish.  I am now committed to clearing old PRs.

I will merge this if the main_module_name is moved to a private, lazy check performed in MakeMaker when the main_module-as-pkg-name does not exist.  i.e., rarely. :)  In other words: only bother looking inside the main module file if Foo/Bar.pm does not exist for dist Foo-Bar.

Tests would be gerat, too, but I can live without them. ;)

rjbs:
Oops, closing was a mistake!

PR #168, opened 2013-04-17
Make Plugin::PkgVersion and perlcritic's RequireUseStrict get along.

hartzell:
This change makes PkgVersion and RequireUseStrict get along.

It is based on Ricardo's suggestion of walking forward across any use statements (PPI::Statement::Include) but goes a bit further to carefully walk out to the end of the line, thereby avoiding separating comments from their commentees.

It has to dance around a bit to handle differences in PPI's insert_after methods, depending on the class of PPI element the insertion point lands on.

All of the original tests pass.  I've added a separate set of tests that take a bigger picture look at the output, so that I could convince myself that everything works.  They use Test::String so that I didn't have to do the monkey work of figuring out where the expected and empirical results differed.  If you give me some feedback I'm willing to try to merge them into the main PkgVersion tests.


hartzell:
@rjbs  ping.  If I were to either take these changes and get them cleaned up w.r.t. the current master, or just take the ideas and do it over, do you think you'll have a chance to give them a review?

rjbs:
Yes, I will merge this kind of change if rebased and/or reworked.

stphnlyd:
@hartzell @rjbs
Have you got a chance to rebase/rework on this?

Without this being fixed, I feel it very inconvenient to use perlcritic with dzil's PkgVersion plugin.

I am bringing this up because it would be not only for my personal usage, but I am looking at promoting dzil in my company, and perlcritic is a key thing that I would like to have with dzil. I would like the settings be as natural and smooth as possible so we could have a better chance to success. 

I know @karenetheridge 's way should work. But AFAIK most Perl users tend to putting "use strict" below the package statement. So it could be inconvenient for them to switch. That OurPkgVersion plugin is an option for me, but it looks somewhat less "standard" as it's outside the default Dist::Zilla package. It got less introduced in various online dzil tutorials, also it requires a "# VERSION" in the source.

hartzell:
I'm not doing much Perl work these days.  @stphnlyd, if you'd like to pick this up and rework it, go for it!

PR #170, opened 2013-04-23
Enables user to configure perl executable used during build and test

mvgrimes:
This patch enables users to configure the which perl executable Dist::Zilla
will use to perform build, test and install actions. It will use the value in
the DZIL_PERL env variable or default to $^X if DZIL_PERL isn't set.

This allows developers to use dzil installed under one version of
perl/local::lib, on projects which require a different version of
perl/local::lib to build/test/install. This has a couple of advantages:

    - you only need to install Dist::Zilla (and plugins and dependencies) once
    - you can keep your active project's installed modules to a minimum, so
      testing dependencies with "rm -rf .local-lib && dzil listdeps | cpanm &&
      dzil test" becomes managable
    - and testing with multiple versions of perl doesn't require a full
      Dist::Zilla toolchain install

If you are receptive to this new config option, I would be happy to update docs
and send pull requests for some of the more popular plugins where $^X is used.
I'm also open to alternative solutions.

I considered implementing as a dzil command line argument, but the env variable
is consistent with local::lib's method of modifying the environment. I don't
believe this should be implemented as a plugin, because it is really developer
specific, not project specific. Most projects expect developers to share a
dist.ini, and if I add

  [PerlExec]
  perl=/home/mgrimes/perl5/perlbrew/perls/perl-5.14-clean/bin/perl

to it is highly unlikely to work for another contributor.

### DIFF
--- a/lib/Dist/Zilla/Dist/Builder.pm
+++ b/lib/Dist/Zilla/Dist/Builder.pm
@@ -631,9 +631,10 @@ sub install {
   eval {
     ## no critic Punctuation
     my $wd = File::pushd::pushd($target);
+    my $perl = $ENV{DZIL_PERL} ? $ENV{DZIL_PERL} : $^X;
     my @cmd = $arg->{install_command}
             ? @{ $arg->{install_command} }
-            : ($^X => '-MCPAN' =>
+            : ($perl => '-MCPAN' =>
                 $^O eq 'MSWin32' ? q{-e"install '.'"} : '-einstall "."');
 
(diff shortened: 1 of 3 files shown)


rjbs:
I have a (more thorough, I think?) implementation of this on my `alt-perl` branch, but I did not merge it because it didn't solve the problems we wanted it to, like testing against many perls, because you can only have the build ready once.

I think the real solution here is a better system for doing multi-perl builds.

PR #174, opened 2013-05-11
move register_stash from Dist::Zilla::MVP::Assembler::Zilla to Dist::Zil...

karenetheridge:
...la, for easier access

This will make what I was trying to do in https://github.com/keedi/Pod-Weaver-Section-Contributors/pull/3 and https://github.com/keedi/Pod-Weaver-Section-Contributors/pull/4 more possible -- it would be somewhat silly to expose the $assembler just to be able to access the stashes that are on $zilla anyway.

### DIFF
--- a/Changes
+++ b/Changes
@@ -4,6 +4,9 @@ Revision history for {{$dist->name}}
           fix docs to not suggest obsolete "Prereq" name (thanks, Ivan
           Bessarabov!)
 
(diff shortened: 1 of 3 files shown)


karenetheridge:
Possibly for v6, the concept of stashes could be expanded, with an API that allows plugins to share data with each other.  e.g. various plugins that deal with git might want to share configs.

PR #189, opened 2013-07-02
Let CPAN::Uploader load the config

doherty:
This lets it handle the newly-supported situation where the `.pause` file is GPG-encrypted.


rjbs:
This change is in place.

PR #197, opened 2013-07-30
travis: build against 5.8 and 5.18

RsrchBoy:
Travis now supports perls 5.8 and 5.18 natively, so let's take advantage of
that :)

Unfortunately, we cannot build against 5.8 at the moment, but 5.8 is
supported.  So... hm.  We'll need a cleverer solution here, but until then
we can mark the 5.8 build as allowed to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,10 @@ perl:
     - "5.16"
     - "5.18"
 
+matrix:
+    allow_failures:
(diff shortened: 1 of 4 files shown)


rjbs:
I've added 5.18 and 5.20 to the .travis.yml

(this PR was kind of messed up, so I didn't use it)

PR #219, opened 2013-10-04
Add 'skip' dist.ini param to PkgVersion plugin

fffinkel:
Allows specification of any path or path part to skip when using PkgVersion to
version files.

Created for the case where a file was too large for PPI::Document.


rjbs:
…and that's my suggestion.  I hope to add a simpler way to one-off file-finders in the future.

PR #226, opened 2013-10-15
FIX: call all builders' build methods to ensure a proper build

karenetheridge:
... just as with -TestRunner plugins, we should call all of them, not just the first.

(This situation arises for me when I use both [MakeMaker::Fallback] and [ModuleBuildTiny] in a dist -- the ->build sub in the former is neutered to not do anything.)


karenetheridge:
rebased to 5.034

rjbs:
thanks, applied, but made ensure_blib private for now

PR #263, opened 2014-01-05
add ->has_encoding & ->is_encoded_content to files

madsen:
This is a followup to #254.  In this version, `has_encoding` acts as a normal Moose predicate (except for `FromCode` files, where it always returns true because the encoding attribute is read-only).  The new method `is_encoded_content` indicates whether the "canonical" version is `encoded_content` or `content`.

If you're happy with this, I'll write up some docs.

The idea is that an EncodingProvider can use `has_encoding` to determine whether it's possible to set the encoding, and then use `is_encoded_content` to decide whether it can examine the file to determine an encoding.

### DIFF
--- a/Changes
+++ b/Changes
@@ -2,6 +2,8 @@ Revision history for {{$dist->name}}
 
 {{$NEXT}}
         - load DateTime lazily
+	- add ->has_encoding & ->is_encoded_content to files
+	  (thanks, Christopher J. Madsen)
 
 5.019     2014-05-20 21:11:47-04:00 America/New_York
         - remove a very brief-lived attempt to double-decode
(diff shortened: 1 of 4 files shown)


rjbs:
I have put this PR off because I've never felt good about the names, nor did I have any better ones to suggest.  Now I suggest this:

File will require:
* has_content
* has_encoded_content
* has_encoding

An auto-detector of encodings can look for:

```perl
if ($file->has_encoded_content && ! $file->has_content && ! $file->has_encoding) { ... }
```

I will code that all up myself if you agree this is still useful and sane, @madsen.

madsen:
That seems like a viable approach to me.  Looks like I'll finally get to release the SmartEncoding plugin I've been sitting on for 2 years :confused:.  Thanks. :smiley:

madsen:
Can we get this into v6 before it leaves trial status?  This sort of API change seems tailor-made for a major-version bump.

PR #267, opened 2014-01-16
use ::File::InMemory in [GenerateFile::Template], to allow later munging

karenetheridge:


rjbs:
I am tentatively rejecting this.  I don't think we want to have the GatherDir::Template plugin acting in multiple phases, and certainly not storing the files between.

PR #286, opened 2014-02-19
Zilla.pm and Builder.pm: Make distmeta fully dynamic and self-inject

kentfredric:
prereqs.

Looks kind crazy, but almost isn't.

Perks: Nobody can modify ->distmeta directly, don't even try.

Downsides: -MetaProvides stuff runs multiple times.

Though for some, this is a **perk**. The problem with metaprovides
being executed when the first instance of distmeta is called, is not
everything in dzil is entirely able to have its chips passed in by then.

Especially eschoteric tricks like trying to record metadata about the
munging of files themselves in MetaJSON.

An alternative approach may be to have a -MetaProvidesOnce role,
  and have a seperate key for that data.

Maybe -MetaProvidesOnceEarly or -MetaProvidesOnceLate to disambiguate
between things that are safe to run before munging, and those that need
munging to occur first.

I don't know, I'm tired and this code feels pretty crazy. But it might
be crazy enough to be useful.

### DIFF
--- a/lib/Dist/Zilla.pm
+++ b/lib/Dist/Zilla.pm
@@ -485,13 +485,20 @@ hash; use a MetaProvider plugin instead.
 
 =cut
 
-has distmeta => (
-  is   => 'ro',
-  isa  => 'HashRef',
-  init_arg  => undef,
-  lazy      => 1,
+has _frozen_distmeta => (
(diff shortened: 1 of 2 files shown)


rjbs:
I don't think this is worth the change.

PR #287, opened 2014-02-22
link to metacpan.org instead of s.c.o for metaspec

karenetheridge:
tests loosened, to accomodate both s.c.o and metacpan in link after running through CPAN::Meta::Converter (see also https://github.com/Perl-Toolchain-Gang/CPAN-Meta/pull/47)

### DIFF
--- a/lib/Dist/Zilla.pm
+++ b/lib/Dist/Zilla.pm
@@ -499,7 +499,7 @@ sub _build_distmeta {
   my $meta = {
     'meta-spec' => {
       version => 2,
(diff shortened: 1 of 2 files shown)


rjbs:
@karenetheridge What's the status?

karenetheridge:
Still stalled.

karenetheridge:
This commit is fine now. The current CPAN::Meta::Spec allows for anything in the 'url' key, so we won't break META checks.

rjbs:
Thanks, applied!

PR #299, opened 2014-03-21
Make PkgVersion skip pod files

dagolden:


rjbs:
Yeah, the finders need wrangling.

I am going to apply this as is, for now, and we can fine-tune it by using a better Finder in the future.  (i.e., it's hardcoded to skip .pod files for now, and we'll make it switchable later by using finders, not options)

PR #310, opened 2014-04-26
default dzil install to cpan instead of cpanm

wchristian:
As of 1.7001 (Sep 08, 2013) cpanm is broken in at least these known ways on
MSWin32:

- a bug that forces system owners to perform many deletion steps when cleaning up .build manually: https://github.com/miyagawa/cpanminus/issues/370
- one bug unrelated to dzil (has to do with carton)
- a bug that causes build.log to replicate and fill up the hdd: https://github.com/miyagawa/cpanminus/pull/368

Despite the latter issue being quite urgent and a fix having been merged, miyagawa doesn't seem to show inclination to actually do a release anytime soon and hasn't communicated any intent to do one despite being asked to.

As such i consider cpanm to be unsuitable for default use by Dist::Zilla.

### DIFF
--- a/Changes
+++ b/Changes
@@ -3,6 +3,8 @@ Revision history for {{$dist->name}}
 {{$NEXT}}
         - NextRelease now bases the new file on disk on the original file on
           disk, skipping any munging that happened in between
+        - default dzil install to use cpan instead of cpanm for higher
+          portability
 
 5.032     2015-02-21 09:36:00-05:00 America/New_York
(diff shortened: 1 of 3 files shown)


rjbs:
I just booted up Win32 and upgraded cpanm with cpanm, successfully.  The bugs all seemed fixed, I think.

I would like to make it possible to pick your default install tool, later, but it doesn't seem pressing right now.

PR #317, opened 2014-05-21
Minor tweaks to Path Handling interfaces to avoid forcing end users to use Path::Class directly

kentfredric:
Attached are some minor tweaks to force Path coercion to Path::Class objects instead of demanding end users to use Path::Class.

- Allows Plugin authors to pass bare strings
- Allows Plugin authors to pass Path::Tiny objects.

These changes should have no impact on *internal* interaction and only make the API more resilient, and help smooth eventual transition to Path::Tiny by allowing more Plugins to use it **now**

/cc @karenetheridge


rjbs:
thanks, merged!

PR #351, opened 2014-09-09
Add --exe flag to authordeps

szabgab:
Make it easy to install dependencies even on operating systemis that don't support piping output
of one command to the input of another command.


rjbs:
I just booted up my Windows machine, ran CMD, and ran:

    C:\>  perl -E "say 'Dist::Zilla'" | cpanm

...and it worked fine.  If you can explain where this doesn't work, please re-open and let me know!

PR #372, opened 2014-11-03
Add the ability to specify the directory to mint new dists in

apocalypse:
I might be zany but I don't like the default naming convention where it uses the dist name as the target directory. For example:

dzil new Foo::Bar # will create Foo-Bar

But I want the dir to be "perl-foo-bar" to differentiate my repos and to reduce the strain on my hands when typing :) I've come up with a solution that isn't invasive and preserves the default. I checked on grep.cpan.me and nothing matches "_mint_target_dir" so I think it's ok to munge it a bit, ha!

In my MintingProfile::Author::APOCAL I'll simply set:

sub mint_dir {
  return 'perl-' . lc( $_[1] );
}

Looks simple, eh? Thanks again for looking into this!

### DIFF
--- a/lib/Dist/Zilla/Dist/Minter.pm
+++ b/lib/Dist/Zilla/Dist/Minter.pm
@@ -70,27 +70,28 @@ sub _new_from_profile {
   );
 
   my $self = $seq->section_named('_')->zilla;
-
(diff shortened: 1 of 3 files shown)


rjbs:
the template dir sounds good to me

PR #377, opened 2014-11-13
Add static_attribution attribution to MakeMaker plugin

dagolden:
This commit adds an option to omit the dzil version number from the
"generated by" comment in Makefile.PL.  For dists that copy the
Makefile.PL back to the repo root, this avoids churn just because dzil
was upgraded.


rjbs:
@dagolden I ignored this commit for a long time, and we had numerous discussions about it and related issues.  What is your current solution to this problem?  Do you think this patch should still be applied?

(I worry that I'll want/need to make it a bit more generic for the sake of other generated-file plugins that either now add such a line or should add one in the future.)

dagolden:
I still think it would be nice to have, but I implemented this in meantime: https://metacpan.org/pod/Dist::Zilla::Plugin::CopyFilesFromBuild::Filtered

Feel free to close this if you prefer.

rjbs:
Merged.  Later I think I'd like to make AddsAttribution a role.

PR #381, opened 2014-11-29
add license module prerequirement to authordeps

zurborg:
In cases where not a regular Software::License module is needed, but another one, dzil will fail directly after ```[DZ] beginning to build Dist-Zilla``` with a ```Can't locate Software/License/XXX.pm``` message. The missing module is *not* listed in ```dzil authordeps```.

The usual workaroud is to explicitly add a ```; authordep Software::License::XXX``` line in dist.ini.

This PR solves this issue by adding the license module to the authordeps list.


rjbs:
Thanks very much, good idea.  Merged!

PR #383, opened 2014-12-01
[NextRelease]: use the original Changes content as a base for munging, rather than whatever is in the repo after release

karenetheridge:
This allows other AfterRelease plugins to copy the Changes file from the built
distribution back to the repo and commit it, before we add in the {{$NEXT}}
line here.

This lets us commit the Changes file at the release tag exactly how it appears in the shipped dist.


karenetheridge:
rebased and commit message shortened.

rjbs:
thanks, applied!

PR #401, opened 2015-01-01
fix [GatherDir] handling of symlinks to directories

karenetheridge:
as seen on irc:

    14:36 < kentnl> ether: yeah, it adds the symlink as a file in addition to following it 
    14:36 < kentnl> ( in the event the symlink is a dir symlink )

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,7 @@
 Revision history for {{$dist->name}}
 
 {{$NEXT}}
+        - fixed [GatherDir]'s handling of symlinks to directories
 
 5.029     2014-12-14 14:44:44-05:00 America/New_York
         - fix new error in [PkgVersion] when a module had no package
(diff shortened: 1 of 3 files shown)


karenetheridge:
Alternatively, that bit of code that checks `->file` or `->symlink` could be turned into this:

```
  if ($self->follow_symlinks) {                                                                                     
    $rule->or(
      # symlinks to files still count as files
      $rule->new->file,
      # traverse into the linked dir, but screen it out later
      $rule->new->symlink,
    );
  } else {                                                                                                          
    $rule->file;
  }
```

It is more efficient at runtime, but a little less readable. I couldn't decide between the two, so I leave it up to you. :)

karenetheridge:
amended with a Changes entry and rebased to master.

rjbs:
This gets my "merged on faith more than careful analysis" stamp.  Thanks!

PR #402, opened 2015-01-01
Make PkgVersion more efficient; pass code snippet as argument

karenetheridge:
Performs all checks once, up front, and passes the $VERSION code snippet as an
argument, to facilitate easier subclassing.


karenetheridge:
If you take #403 first, I can clean this up after (as it happens, with the our-without-braces option being added, I no longer need to call methods in this plugin directly - but I think it's still worth keeping the changes that do the checks up front just once.)

karenetheridge:
The changes here would still be useful, but I reworked other things to not block on this.

PR #403, opened 2015-01-04
PkgVersion.version_format enum

karenetheridge:
As discussed on irc.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,8 @@
 Revision history for {{$dist->name}}
 
 {{$NEXT}}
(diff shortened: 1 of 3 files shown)


rjbs:
Thanks, but this isn't quite what I want to do.  I think (but want a bit more thinking) that what we want is to leave PkgVersion as it is (although I've deprecated use_our) and to add a ReplacePackageDeclaration-like plugin, so you can replace `package X::Y;` with `package X::Y; $VERSION = 1.2.3` or something

but I wanna think a bit more

PR #405, opened 2015-01-05
find and filter out all namespaces found in scanned modules

### LINKED ISSUE #404: AutoPrereqs fails to account for non-primary packages in a file and happily creates self-dependencies
XML::Builder 0.903 is broken because of `[AutoPrereqs]`.

The distribution contains [a single `.pm` containing all of its classes](https://github.com/ap/XML-Builder/blob/v0.903/lib/XML/Builder.pm), some of which `use parent -norequire` others. Since AutoPrereqs thinks it should list any package that appeared in a `use parent` line as a runtime prereq, and doesn’t notice that some of them are also *provided* by the very same distribution (nay, the very same *file*), this [leads it to declared these ISA relationships as self-dependencies](https://metacpan.org/source/ARISTOTLE/XML-Builder-0.903/META.json).

[The `dist.ini` involved](https://github.com/ap/XML-Builder/blob/v0.903/dist.ini) is a rather vanilla affair.

(Reported as a bug as per @karenetheridge’s opinion. In the meantime I will be working around it by setting `skip = ^XML::Builder($|:)` on AutoPrereqs.)

karenetheridge:
closes #404.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,8 @@
 Revision history for {{$dist->name}}
 
 {{$NEXT}}
+        - [AutoPrereqs] now filters out all namespaces found in contained
+          modules, not just the one corresponding to the module filename
 
 5.029     2014-12-14 14:44:44-05:00 America/New_York
         - fix new error in [PkgVersion] when a module had no package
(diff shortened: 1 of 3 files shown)


rjbs:
Excellent, thanks!  Merged.

PR #408, opened 2015-01-11
use canonical forms for RT links in metadata

karenetheridge:


dolmen:
Note to self: many other RT links to fix: http://grep.cpan.me/?q=http%3A%2F%2Frt.cpan.org%2FNoAuth%2FBugs.html

rjbs:
Thanks, applied!

PR #409, opened 2015-01-11
be a bit more selective when scanning for local namespaces

karenetheridge:
This screens out things like "t::00-report-prereqs.t" (translated from
filenames) and '#' (from "package # hide from PAUSE").

(a follow-up from the change to AutoPrereqs in 5.030.)

### DIFF
--- a/corpus/dist/AutoPrereqs/lib/DZPA/Main.pm
+++ b/corpus/dist/AutoPrereqs/lib/DZPA/Main.pm
@@ -49,6 +49,9 @@ use parent -norequire => 'DZPA::Second';
 
 use DZPA::Empty;
 
+package # hide from PAUSE
+    DZPA::Fourth;
+
 
 __END__
 =head1 FOO
--- a/lib/Dist/Zilla/Plugin/AutoPrereqs.pm
+++ b/lib/Dist/Zilla/Plugin/AutoPrereqs.pm
@@ -183,6 +183,7 @@ sub register_prereqs {
       }
       s{\.pm$}{} for @this_thing;
       s{/}{::}g for @this_thing;
+      @this_thing = List::MoreUtils::uniq grep { /^\w+(?:(?:'|::)\w+)*$/ } @this_thing;
 
       # this is a bunk heuristic and can still capture strings from pod - the
       # proper thing to do is grab all packages from Module::Metadata
@@ -208,7 +209,8 @@ sub register_prereqs {
     }
 
     # remove prereqs shipped with current dist
-    $self->log_debug([ 'excluding local packages: %s', sub { join(', ', List::Util::uniq @modules) } ]);
+    @modules = List::Util::uniq(@modules);
+    $self->log_debug([ 'excluding local packages: %s', sub { join(', ', @modules) } ]);
     $req->clear_requirement($_) for @modules;
 
     $req->clear_requirement($_) for qw(Config DB Errno NEXT Pod::Functions); # never indexed


rjbs:
```
+ @this_thing = List::MoreUtils::uniq grep { /^\w+(?:(?:'|::)\w+)*$/ } @this_thing;
```

I'm not 100% sure I'm clear on your intent, so bear with me, here.

I think your first grep is there to limit `@this_thing` (which is possible package names for the file under consideration) to valid package names.  So, for example, not `foo::bar.t`.  I think there are two better ways to do this, and I can switch it to either one if you agree.  Either (1) after the `s/.pm$//` and the slash-to-:: conversion, grep out invalid module names (using the ModuleName type) or (2) don't consider putting into `@modules` anything that didn't originally end in `.pm`.  Or a combination of both.

```
- push @this_thing, $file->content =~ /package\s+([^\s;]+)/g;
+ push @this_thing, $file->content =~ /(?!#)\S*package\s+(\w+(?:(?:'|::)\w+)*)/g;
```

You indicated that this change it to skip `package # hide from pause \n Whatever::YourFace`.  Does it actually do that?  I don't think that regexp does what you I think you think it does.  But without a test, I don't know which one of us is right.

I think a better test might be to check whether there is anything matching `/\H/` between package and the package name.

Now, all that said:  why bother?  The packages being gathered there are to find ones to exclude from prereqs.  If it sees a "hidden" package called Foo::Bar, that will tell it to ignore `use Foo::Bar` elsewhere, because the dist is shipping it.  Are you *specifically* concerned with the case that you `use` Foo::Bar, and rely on it coming from CPAN and being in prereqs, but *also* have a "hidden" declaration of it in your code?  I think that relying on the "hide from PAUSE" mechanism might be confusing two things, here, and that you're better off specifying a manual prereq in this situation, but I'm open to argument.  Or at least explanation of whether I understood!

karenetheridge:
ok, lots of things going on here:

- using the ModuleName type wouldn't be correct, as its regex has been wrong since 5.16 (but that's a separate issue, and if we can get that fixed, then my objections here are moot of course)
- writing tests is difficult, as what would be getting tested?  the outcome is *usually* the same - e.g. we don't see '# hide from PAUSE' appearing in prereq specifications today.
- I'm not sure what you're asking about regarding the regex. but here:

    perl -wle'my $content = "foo bar package # hide from PAUSE";  print $content =~ /(?!#)\S*package\s+(\w+(?:(?:::)\w+)*)/g'

- my intent was mostly to improve efficiency (and the amount of things dumped during debugging) by excluding not-actually-a-package things earlier in the scan.  The code has *never* found package names that have an embedded `# hide from PAUSE`, and I am not trying to change that here.

rjbs:
I don't have any bug reports on MooseX::Types::Perl, or they'd be fixed.  Got details?

I think I'm confused because I think you're giving two distinct changes for two distinct reasons in one commit with one sentence of explanation.

So:

* the first line added is to skip filenames that don't end up becoming useful packages
* the second one is to skip commented-out package lines *and* package-followed-by-non-package-name?

I'm confused by the regex because you added `(?!#)` and then your commit message talks about "# hide from PAUSE", which is, as you say, not affected by the lookahead.  The thing that's making it skip that case faster is the package-like bit *after* `package`.  But if you want to skip commented-out package lines, it's not a good regex either, because of the `\S` etc.

I don't need a test to merge this, but I need to understand the intent, because as far as I can tell, the regexp isn't quite right for any intent I've guessed at.  I am not going to ask, either, for further revisions or whatever from *you*.  I just want to see what specifically these changes were for, so I can tell whether I want to change the regexp before applying.

"prune more stuff earlier" makes sense.  Matching package names with a package name matcher makes sense.  Then I'm just left wondering about the lookahead.  Is that, in fact, meant to be filtering out commented-out package statements?

I think that's my only question.  Everything else above is maybe just there so you can tell me I've totally missed the point.

karenetheridge:
> I don't have any bug reports on MooseX::Types::Perl, or they'd be fixed. Got details?

That's the unicode-in-module-names issue; I thought I'd filed tickets against everything relevant (this is stuff that came out of Acme::LookOfDisapproval, and my notes are stuck on my rescued HD, not here right now - will follow up).

> Then I'm just left wondering about the lookahead. Is that, in fact, meant to be filtering out commented-out package statements?

Ah, I understand now why you were confused. Yes, there are two things I was going for here, and they ought to have been in separate expressions to properly achieve them:  1, skip everything after a `#`, and 2. filter out things that can't be package names.

sorry for all the confusion. this ought to have been a simple change, but I muffed it by trying to do too much at once.

karenetheridge:
okay so I rebased this PR now that #454 is in -- the only relevant bits that remain are:

```
+      @this_thing = List::MoreUtils::uniq grep { /^\w+(?:(?:'|::)\w+)*$/ } @this_thing;
```

and
```
-    $self->log_debug([ 'excluding local packages: %s', sub { join(', ', List::MoreUtils::uniq @modules) } ]);
+    @modules = List::MoreUtils::uniq @modules;
+    $self->log_debug([ 'excluding local packages: %s', sub { join(', ', @modules) } ]);
```

Also, the exclusion of this isn't being tested, so that could be added too:

```
package # hide from PAUSE
    DZPA::Fourth;                                                                                                   
```

But most of my intent in thsi PR has already been resolved, so.. meh?

rjbs:
Thanks, applied!

PR #410, opened 2015-01-14
Emit dist_ini()s generated text with payload keys sorted.

kentfredric:
This makes the output format more stable, in the event that anything
needs to round-trip those files through some process and recompare them.

For instance, in the case of bakeini, it retains input order and only expands bundles.

Thus, testing bakeini is made more complicated by dist_ini() having a
random order of each payload key, that can only otherwise be tested by a
full parse instead of a simple line-by-line comparison.

### DIFF
--- a/lib/Test/DZil.pm
+++ b/lib/Test/DZil.pm
@@ -113,7 +113,7 @@ sub _build_ini_builder {
 
     my $config = '';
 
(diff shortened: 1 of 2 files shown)


karenetheridge:
looks sane to me :)

rjbs:
thanks, applied

PR #411, opened 2015-01-25
eliminate giant swathes of whitespace in generated Makefile.PL

karenetheridge:
because handsome is as handsome does


rjbs:
you have successfully appealed to my vanity.

PR #417, opened 2015-02-05
Documented 'version' attribute in Dist::Zilla::PluginBundle::Filter

mjemmeson:
Noticed this wasn't documented (and no comment in code to suggest not to use it)

thanks,
Michael


rjbs:
Thanks, applied!

PR #420, opened 2015-02-08
follow Moose best practices by method-modifying the existing BUILDARGS

karenetheridge:


karenetheridge:
rebased to 5.036

rjbs:
I am just barely vaguely uneasy about this, and wonder whether I will be sad later that we didn't get conflict in some case where we could have…

…but I'm applying anyway, because because.

Thanks, applied!

PR #422, opened 2015-02-18
document use of Encoding to avoid problems with AutoPrereqs

dagolden:
Inspired by this blog post:
[Non-obvious fix to a dzil problem](http://goo.gl/rsFQXd)


dolmen:
:+1:

But add a `L<>` link to the Encoding plugin for easy browsing to its documentation.

dagolden:
@dolmen, done, squashed and pushed

rjbs:
Thanks, I saw the blog post and had the same idea!

rjbs:
i was thinking I'd like to go even further and provide a pointer in the error messages.  I'll look at doing that, too.  Meanwhile: Thanks, applied!

jmarca:
So succinct!  Awesome 9 lines of explanation encapsulates exactly what I was getting at in my meandering blog post.

PR #424, opened 2015-02-21
main_module lookup improvements

dolmen:
Improvements of the `main_module` lookup:
* show the reason why the guessing failed even if a .pm exists outside `lib/`
* refactoring: faster guessing
* fix indent around previous changes in otherwise untouched code


rjbs:
Thanks!

I've merged the check fix (adding ^lib) and made indent fixes, but the 16 line increase did not strike me as worth it.

PR #428, opened 2015-02-27
add use_begin option in the PkgVersion plugin, to re-enable BEGIN { $VERSION= }

dams:
commit 62e97df removed the feature of having VERSION set in a BEGIN block. It turns out that some people *do* use Dist::Zilla to build XS packages, and they need XSLoader to happen at BEGIN time, with VERSION already set.

This PR is for instance required for the following re-engine-pcre PR:  https://github.com/avar/re-engine-pcre/pull/2

Thanks !

### DIFF
--- a/lib/Dist/Zilla/Plugin/PkgVersion.pm
+++ b/lib/Dist/Zilla/Plugin/PkgVersion.pm
@@ -65,6 +65,13 @@ with some analyzers.  Use of this feature is deprecated.
 
 Something else will replace it in the future.
 
+=attr use_begin
+
+The idea here is to wrap the version assignment in a BEGIN bloc. This helps
+when using dist.ini in distribution that contains XS code, and where DynaLoader
+has t be called at BEGIN time, and requires VERSION. Yes, there are intrepid
+heroes that are using Dist::Zilla with XS code. Defaults is false.
+
 =attr finder
 
 =for stopwords FileFinder
@@ -117,6 +124,12 @@ has die_on_line_insertion => (
 );
 
 has use_our => (
+  is  => 'ro',
+  isa => 'Bool',
+  default => 0,
+);
+
+has use_begin => (
   is  => 'ro',
   isa => 'Bool',
   default => 0,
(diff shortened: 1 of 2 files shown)


karenetheridge:
Instead of doing this, what is wrong with doing what other dzilified XSLoader-using modules do (for example Class::Load::XS) - https://metacpan.org/source/ETHER/Class-Load-XS-0.09/lib/Class/Load/XS.pm#L11 --

```
use XSLoader;
XSLoader::load(
    __PACKAGE__,
    exists $Class::Load::XS::{VERSION}
        ? ${ $Class::Load::XS::{VERSION} }
        : (),
);
```

karenetheridge:
Is the XSLoader call truly needing to be in a BEGIN block? I don't see anything in the code that requires it.

avar:
Maybe it's not. I haven't looked into the PCRE dist for a while, and I couldn't compile it now when I tried due to changes in the pcre API.

However given that 62e97df guards against:

    package Blah;
    our $VERSION = 1.23;
    use NonExisting;
    1;

But not:

   package Blah:
   our $VERSION = 1.23;
   use Existing;
   my $thingy = Existing::nonexisting();
   1;

And that 5.8 is ancient now I think it makes sense as a general fix to revert 62e97df, it's more idiot-proof that using $VERSION in a BEGIN block Just Works with Dist::Zilla by default than us trying to work around some obscure 5.8 bug.

karenetheridge:
IMHO, distributions where a plain old `our $VERSION = '1.23';` isn't suffficient should manage their version through some other mechanism than [PkgVersion].  I recommend [RewriteVersion] and [BumpVersionAfterRelease].

avar:
But why? If the concern isn't obscure 5.8 bugs setting the version in a
BEGIN block is better without any downsides.

avar:
To add to that a bit, since I was on mobile at the time, I mean that if setting the version in a BEGIN block is categorically going to work for more code written for non-deprecated perls than setting it outside a BEGIN block then doing the former seems like a saner default.

rjbs:
Thanks, I have applied this PR with some amendments.  I would *really* love a test case that didn't violate the plugin's encapsulation, though.  Anybody?

karenetheridge:
I think this can be closed now?

rjbs:
@avar I may well change the default in the future, because I think the BEGIN block is better, but for now, I'm leaving it off for the sake of not changing the deployed behavior.

PR #429, opened 2015-03-01
add 'use warnings' to D-Z-App-Tester per CPANTS

andyjack:


rjbs:
Nice catch, thanks.  Merged!

PR #430, opened 2015-03-01
build prep: add error checking and handle non-blocking rmdir on Path::Class::Dir::rmtree

wchristian:
Oh jeeze, forgot the description.

For one, rmtree in that code was completely unchecked. This adds a check in a commit that's likely to be completely uncontroversial.

For the other, as covered in https://rt.perl.org/Ticket/Display.html?id=123958 , rmdir and thus rmtree are non-blocking on windows. This adds code to handle that gracefully with a 1-2 second cpu spin.


rjbs:
thanks, applied!

PR #432, opened 2015-03-03
Mailmap rwstauner's cpan address

rwstauner:


rjbs:
merged with pleasure, thanks

PR #433, opened 2015-03-07
detect when a plugin in ~/.dzil/config.ini is not installed

karenetheridge:
Presently, all failures are greeted with "Your global configuration file
couldn't be loaded...", which is not always accurate.


karenetheridge:
rebased to 5.036

rjbs:
thanks, applied!

PR #434, opened 2015-03-08
0.040 can now parse "Required plugin bundle ... isn't installed"

karenetheridge:


karenetheridge:
I've rebased this to 5.033, and added another commit which bumps up the prereq on Config::MVP to get us past the broken section_name interface in ::WithBundles.

karenetheridge:
one more commit added - this tests the bad interaction with Config::MVP 2.200009 (showing the prereq bump is needed), and also fixes the regexp that detects whether it was a bundle or a plugin that is problematic.

rjbs:
merged, thanks

PR #438, opened 2015-03-29
Make release status configurable

dagolden:
This change introduces a new attribute, 'release_status', and a new
role, D::Z::R::ReleaseStatusProvider.

The 'is_trial' variable is changed to be read-only.  This will break the
handful of CPAN modules that try to set it.  Those modules should set
the RELEASE_STATUS environment variable before building instead.

For backwards compatibility, 'is_trial' in dist.ini and the TRIAL
environment variable are supported.

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -54,6 +54,9 @@ remove = Some::Package::That::Does::Not::Exist::Due::To::A::Typo
 [Breaks]
 Dist::Zilla::Plugin::MakeMaker::Awesome = < 0.22
 Dist::Zilla::App::Command::stale = < 0.040
+Dist::Zilla::Plugin::TrialVersionComment = <= 0.003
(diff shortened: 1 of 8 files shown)


dagolden:
I still need to send patches to the affected distributions, so please
don't ship this yet.

dagolden:
OK.  All dists that I could find via grep.cpan.me have pull requests, so I think this is ready to go once you've reviewed it.

karenetheridge:
If there is any runtime code affected, could `x_breaks` metadata be added for those modules?

dagolden:
@karenetheridge, done

tsibley:
::update 0.05, just released, ensures compat with this PR.

rjbs:
Thanks, 5.035 released!

PR #443, opened 2015-04-11
Test::DZil: Load Dist::Zilla::Tester on demand

kentfredric:
Although most people who load this module will be wanting the builder
functions, there are a couple of utility function in it ( namely,
simple_ini ) that are useful to have, even when not invoking main
Dzil code.

Subsequently, if you are testing an App::Command that doesn't normally
need to load dzil, but does some kind of processing on its own of
dist.ini ( Like authordeps does ), then you must presently pay the moose
price to do so.

This minor change avoids that penalty.

The one downside it does have is some people may have been inherently
dependending on DZT being loaded for them ( as was the test ), but that
is reasonably dodgy in itself.

### DIFF
--- a/lib/Test/DZil.pm
+++ b/lib/Test/DZil.pm
@@ -3,7 +3,6 @@ use warnings;
 package Test::DZil;
 # ABSTRACT: tools for testing Dist::Zilla plugins
 
-use Dist::Zilla::Tester;
 use Params::Util qw(_HASH0);
 use JSON::MaybeXS;
 use Scalar::Util qw(blessed);
@@ -38,6 +37,18 @@ Test::DZil provides routines for writing tests for Dist::Zilla plugins.
 These return class names that subclass L<Dist::Zilla::Dist::Builder> or
 L<Dist::Zilla::Dist::Minter>, respectively, with the L<Dist::Zilla::Tester>
 behavior added.
+
+=cut
+
+sub Builder {
+  require Dist::Zilla::Tester;
+  Dist::Zilla::Tester::builder();
+}
+
+sub Minter {
+  require Dist::Zilla::Tester;
+  Dist::Zilla::Tester::minter();
+}
 
 =func is_filelist
 
--- a/t/tester-demo.t
+++ b/t/tester-demo.t
@@ -21,7 +21,7 @@ ok(
 );
 
 ## SIMPLE TEST WITH DZIL TESTER
-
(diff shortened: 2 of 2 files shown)


karenetheridge:
+1

dolmen:
:+1: (because I already had to debug some dzil tests, and asked myself "who is loading Dist::Zilla::Tester?"; I prefer explicit loading)

But before merging we will first have to at least fix tests that rely on the implicit loading.
One example in the wild: https://metacpan.org/source/DAGOLDEN/Dist-Zilla-Plugin-CheckExtraTests-0.026/t/checkxt.t#L1
The query I used (includes many false positives): http://grep.cpan.me/?q=file%3A.t+Dist%3A%3AZilla%3A%3ATester

karenetheridge:
I checked all my distributions, and every test loads `Test::DZil`, which in turn loads `Dist::Zilla::Tester`.

karenetheridge:
Since this is a change that will affect a lot of things downstream, I would request that this change be released as -TRIAL first, to give us a chance to fix tests before distribution installation is affected.

dolmen:
We could add a compatibility shim in `Test::Dzil`:

```perl
unless ($INC{'Dist/Zilla/Tester.pm'}) {
    *Dist::Zilla::Tester = sub ()
    {
        require Dist::Zilla::Tester;
        'Dist::Zilla::Tester'
    }
}
```

karenetheridge:
I tested all my distributions against the change, and only `[PromptIfStale]` needs a release. I haven't yet tested all other plugins I use, though, which I plan to do today when I try to install everything on 5.21.11.

karenetheridge:
(after a PromptIfStale release) everything used by my author bundle installed cleanly under this patch, and so does everything else brought in by my plugin bundle. I'm withdrawing my request for a -TRIAL first.

karenetheridge:
@kentfredric Dist::Zilla::Plugin::Meta::Contributors's tests has issues; can you patch it?

kentfredric:
@karenetheridge, I thought you already did? 

https://github.com/dagolden/Dist-Zilla-Plugin-Meta-Contributors/pull/3

karenetheridge:
lol, never mind then.. :)

rjbs:
Will merge in v6.

rjbs:
Applied to dev-v6 branch.

Issue #363, opened 2014-10-26
t/plugins/manifest.t fails on MSWin32 and cygwin


plicease:
Looks like it is trying to create files with `\` in them, which are of course illegal in windows.  The easy fix would be to skip the test on `$^O =~ /^(MSWin32|cygwin)$/`.  The correct fix might be to write a version of the test that works on a windows file system.

Actually I think the `\` is getting translated to `/`, but the corresponding directives are not there

Error sysopen on '/home/ollisg/dev/Dist-Zilla/tmp/5x4UK3OHTj/source/file'with'quotes/or/backslash.txt63244114575820': No such file or directory

(ie `file'with'quotes` and `file'with'quotes/or` would need to be created before you can write to a file in that directory)

example cpan testers failure:

http://www.cpantesters.org/cpan/report/3ab9f191-6d62-1014-9f05-46b7903388cd

PR #452, opened 2015-05-12
additional fix for #361 on cygwin

plicease:
I reported this as #363 and it was marked a duplicate of #361, but the latter was only fixed on `MSWin32`, not `cygwin`.  This fixes it on Cygwin.


rjbs:
thanks, applied!

PR #454, opened 2015-05-16
tighten up the regex that extracts local packages in AutoPrereqs

karenetheridge:
Skip all comments, words with sigils.

However, we still extract content from pod. The only real remedy for that is to use Module::Metadata. To make that more efficient, I recommend pulling Dist::Zilla::Role::ModuleMetadata into core (which caches the parsed document, just like the PPI role does, for sharing between all plugins that are interested in its data. I currently use three that do this:

https://metacpan.org/requires/distribution/Dist-Zilla-Role-ModuleMetadata?sort=[[2,1]]&size=100


karenetheridge:
The real solution, using MMD:  `push @this_thing, $self->module_metadata_for_file($file)->packages_inside;`

rjbs:
thanks, applied

Issue #442, opened 2015-04-11
[MakeMaker] needs special handling for ranged prereqs


kentfredric:
As per https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/issues/215 , EUMM prior to 7.05_01 responds to ranges in PREREQ_PM by treating them as "0", while spitting a warning about invalid versions.

I've hand-coded myself a workaround for my specific single atom case here, https://github.com/kentnl/Dist-Zilla-Util-ExpandINI/blob/0.003001/Makefile.PL#L66-L93 , but it is more work than it needs to be by far.

The easiest option is to depend on a newer EUMM when such an atom is seen.

But coding around it is somewhat plausible, albeit prone to a few cycles of bug finding.

karenetheridge:
My preference would be to inject a configure-requires prereq on EUMM as appropriate, and let the author decide whether that's worth it to have the version range in their prereqs or not.  We can't do that until EUMM 7.06 is out though of course :)

karenetheridge:
It turns out I had a ranged prereq in Dist::Zilla::Plugin::PromptIfStale 0.041... when I tried to install it in perl 5.21.11, I was greeted with:

    Unparsable version '>= 4.300039, != 5.033' for prerequisite Dist::Zilla at Makefile.PL line 131.

So regardless of what EUMM does, we should handle this in dzil. We should probably at least warn to let the author know that htey're doing something that the toolchain won't like, but bumping up the prereq on EUMM is reasonable too (although the author might not want that prereq, so a warning is a good idea for that too.)

rjbs:
I prefer to wait until we have a stable EUMM to take further action here.

karenetheridge:
EUMM 7.06 is out. Now is the time to discuss the way forward.

karenetheridge:
This is now implemented in https://github.com/rjbs/Dist-Zilla/pull/527.

PR #455, opened 2015-05-21
warn the user if passing version ranges to EUMM, which are treated as 0

karenetheridge:
addresses #442. See also https://github.com/Perl-Toolchain-Gang/ExtUtils-MakeMaker/issues/215

I've released the same change in [MakeMaker::Awesome] version 0.34.


rjbs:
thanks, applied!

PR #456, opened 2015-05-21
remove all uses of List::MoreUtils

karenetheridge:
~~Now only uniq() remains.~~

All uses of List::MoreUtils have been removed.


karenetheridge:
In some other distributions I've been using this instead (note ordering is not preserved) -- this might be preferable to adding a new dependency on List::UtilsBy:

```
sub _uniq { keys %{ +{ map { $_ => undef } @_ } } }
```

rjbs:
Thanks but I'm happy with my current long-distance provider.

(It won't change my mind, but I hope you can find a better way to replace `uniq`.  That one incurs at least N subroutine calls for N inputs, which is pretty lousy if it's just for identity.)

karenetheridge:
I shall revisit when List::Util::uniq exists!

PR #461, opened 2015-06-11
Fix warning on Perl 5.8.8 that causes test failure in file.t

plicease:
Weird but true: on Perl 5.8.8, this gives a warning:

 perl -e 'use warnings; local $ENV{FOO} = undef' ;
 Use of uninitialized value in scalar assignment at -e line 1.

Only seems to happen for `%ENV`.  This warning is not generated on 5.8.9 or 5.10.0.  Because t/file.t is using `Test::FailWarnings` this causes it to fail.  A more elegant refinement of the code might be more appropriate or requiring Perl 5.8.9.

### DIFF
--- a/lib/Dist/Zilla/Tester.pm
+++ b/lib/Dist/Zilla/Tester.pm
@@ -274,7 +274,9 @@ sub minter { 'Dist::Zilla::Tester::_Minter' }
 
     local @INC = map {; ref($_) ? $_ : File::Spec->rel2abs($_) } @INC;
 
-    local $ENV{DZIL_GLOBAL_CONFIG_ROOT} = $tester_arg->{global_config_root};
+    local $ENV{DZIL_GLOBAL_CONFIG_ROOT};
+    $ENV{DZIL_GLOBAL_CONFIG_ROOT} = $tester_arg->{global_config_root}
+      if defined $ENV{DZIL_GLOBAL_CONFIG_ROOT};
 
     my $global_stashes = $self->_setup_global_config(
       $tester_arg->{global_config_root},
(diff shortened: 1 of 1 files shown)


karenetheridge:
I recall a thread on p5p last year about how the undefined value in environment variables was not a good idea, as it doesn't have consistent behaviour everywhere.  More correct is to do: `delete $ENV{FOO}`

plicease:
@karenetheridge updated.  I had to look up the correct incantation for deleting a localization of a hash entry (according to delete in perldoc).  I vaguely remember some hubub about undef in `%ENV` but not the details, but assuming that `delete local` works on `%ENV` as it does with other hash then it should be good?

Verified that it still works on 5.8.8.

Leont:
> I recall a thread on p5p last year about how the undefined value in environment variables was not a good idea, as it doesn't have consistent behaviour everywhere. More correct is to do: delete $ENV{FOO}

AFAIK, a simple «local $ENV{FOO}» should also do the trick

plicease:
I can take out the delete if that is the case.

plicease:
Though it does seem to leave an undef in `%ENV` if that is indeed a problem:

```
twin% perl -MData::Dumper -e 'local $ENV{FOO}; print Dumper(\%ENV)'|grep FOO
          'FOO' => undef,
```

plicease:
I removed the delete (back to my original PR essentially) because:

 - My goal was to replicate the existing behavior of the code, minus the warning
 - I'm not sure what the issue is with `undef` and `%ENV`, but the existing code would have definitely set an environment var to undef (and thus is a perhaps related but separate issue)
 - guessing how to fix problems I can't recall the details to is problematic
 - I did try to break things with `undef` and `%ENV` on Windows and cygwin.

That said adjust as appropriate if details are revealed.

rjbs:
Thanks, applied as is!

rjbs:
…and then fixed.  This essentially did:

  local $x = $y if defined $x;

Clearly it should have been `if defined $y`

Fixed thusly, more or less.

PR #464, opened 2015-06-15
Add a copy parameter to GenerateFile

Altreus:
This is more of a proposal than a PR at the moment, because obviously there's not even a test for it yet.

I thought it would be useful to have a copy option in GenerateFile. My use case is that I'm making a profile that would want to copy more files into the lib/ directory, meaning the paths to the files would want to be templated as well as the content of them.

Rather than listing a bunch of lines, it seemed prudent to do something like this.

### DIFF
--- a/lib/Dist/Zilla/Plugin/GenerateFile.pm
+++ b/lib/Dist/Zilla/Plugin/GenerateFile.pm
@@ -2,6 +2,7 @@ package Dist::Zilla::Plugin::GenerateFile;
 # ABSTRACT: build a custom file from only the plugin configuration
 
 use Moose;
+use Path::Tiny;
 with (
   'Dist::Zilla::Role::FileGatherer',
   'Dist::Zilla::Role::TextTemplate',
@@ -63,6 +64,20 @@ has filename => (
   required => 1,
 );
 
+=attr copy
+
+The C<copy> attribute defines a filename whose lines will be used as
+the C<content> attribute. The lines will be prepended to the C<content>
+arrayref.
+
+=cut
+
+has copy => (
+  is  => 'ro',
+  isa => 'Str',
+);
+
+
 =attr content
 
 The C<content> attribute is an arrayref of lines that will be joined together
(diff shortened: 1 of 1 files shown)


Altreus:
OK I updated the PR with actually-tested code. I've not looked at writing an actual test in `t/` for it yet, mind.

rjbs:
Hi!  Sorry for the delay.  I'm working through my backlog again.

I'm not entirely sure I understand the problem you want to solve.  Is there a reason that GatherDir::Template won't work?

At any rate, I do not like that this PR leaves it possible for both content and copy to be provided, with content being, it seems, silently ignored when copy is set.

Let me know what you want to achieve and maybe I'll be enlightened, or I can help find a way that I think is more suitable for the capability to be added.  Thanks!

karenetheridge:
My guess is he wants to (optionally) provide the file content in an existing file, rather than inlined in dist.ini?  In that case, `[GatherDir::Template]` might be closer to what you need (it doesn't have an option to template the filename yet though).

Altreus:
Yes, I wanted to template the file *and* the filename - and `[GenerateFile]` seemed to be the only one that supported a templated filename.

Supplying the content inline in `dist.ini` just became really cumbersome, once the file became longer than a trivial skeleton, so I thought it would be easier to generate a file based on a different file.

I recognise that this probably means I should try to fix our framework to have less boilerplate in its modules this stage it's easier to make a minting profile for it and ignore the problem!

I didn't consider adding this feature to `[GatherDir::Template]`, mostly because I didn't know about it.

rjbs:
Okay!  I would prefer that GatherDir::Template get the change.  I'm not 100% sure how I'd suggest to implement it, but maybe something like:

```ini
[GatherDir::Template]
; will find file ./the-FOO-vBAR.txt
rewrite.FOO = $zilla->expression->whatever
rewrite.BAR = $]

Altreus:
Done in GatherDir::Template. I decided "rename" made more sense

https://github.com/rjbs/Dist-Zilla/pull/508

PR #465, opened 2015-06-27
omit configure-requires prerequisites from merged_requires

karenetheridge:
This data went into the fallback prereqs used by [MakeMaker], but it's already
too late to fulfill configure prereqs that weren't already in META.

(This is intended to make the ::Fallback behaviour better on older toolchains. cc @ribasushi, @haarg


haarg:
+1

I don't see the value in including configure requirements in the runtime prereqs for older EUMM versions.  The one argument in favor seems to be for dists that only list them there but then assume they will be available in the build/test phases.  I think for Dist::Zilla though requiring people to fully state their prereqs is fine.

rjbs:
Thanks, applied!

PR #466, opened 2015-06-27
[PkgVersion]: add safe $VERSION when version has an underscore

karenetheridge:
Adds $VERSION = eval $VERSION; when using an underscore trial version; it turns out that Dist::Zilla is quite happy to deal with underscore versions everywhere; we just usually don't :)

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,8 @@
 Revision history for {{$dist->name}}
 
 {{$NEXT}}
+        - [PkgVersion] adds "$VERSION = eval $VERSION" when version contains
+          an underscore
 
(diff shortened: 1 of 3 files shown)


karenetheridge:
note: these commits aren't squashed - I kept them separate to reflect our changed consensus after irc discussion. when squashed it's pretty short and I hope sensible.

rjbs:
I'd like to see this use an extra blank line if needed… or something.  It undercuts a but the ability to maintain line numbers.  If we can't do that, this is okay, but… bleh.

I do not like underscores. 😟

rjbs:
While noting that I still *loathe* low lines in versions, I have merged this.  Thank you, I think. :)

PR #467, opened 2015-06-28
Keep build output sorted

karenetheridge:
sorts the dumping of various arrayrefs in Makefile.PL and metadata, for greater consistency between builds. This keeps ad-hoc diffs smaller, and helps distribution packagers.


karenetheridge:
It looks like there is some overlap with #482, but the changes to [ModuleBuild] there are extra.

rjbs:
Thanks, applied.

Issue #292, opened 2014-03-06
make PodCoverage tests an author test


xenoterracide:
I would like PodCoverage tests to be either an `author` test by default, or at least configurable to be one. Motivations is that if someone ads a feature they should add documentation at the same time, instead of leaving it to the maintainer to do, and contributors should not be required to run more than `dzil test`.

PR #468, opened 2015-06-28
Make PodCoverage and PodSyntax tests author tests instead of release tests

autarch:


karenetheridge:
closes #292

autarch:
See http://blog.urth.org/2015/06/28/author-versus-release-tests-with-and-without-distzilla/ for some context of why this makes sense (to me and @karenetheridge, at least ;)

dolmen:
From your blog post your rationale *seems* to be that your release tests do dangerous things and that you don't want to run them from Travis-CI.
On my side, I see no problem in running both author and release tests from Travis-CI. But I see a problem in running PodCoverage and PodSyntax tests in author tests which are run during `dzil test`: it slows down the builds for all dzil users, while I think that POD is not critical until release.

autarch:
It's not so much that they do dangerous things as they check things that I may not want checked between releases. The example in the blog post was a test to check that the Changes file has content. It's easy to imagine pushing changes between releases that don't need a Changes update (like fixing typos in comments).

dolmen:
So it's easy to imagine that fixing typos in comments don't require a run of PodCoverageTests and PodSyntaxTests.

The policy for running PodCoverageTest/PodSyntaxTests should be project dependent (I suggest adding a "phase" parameter to this plugin, with 2 possible values : "author", "release"), but the default staying a "release" test.

karenetheridge:
> So it's easy to imagine that fixing typos in comments don't require a run of PodCoverageTests and PodSyntaxTests.

These tests are still expected to *pass* on every commit. "has content in Changes" is not.  What we're trying to do is separate the extra tests into categories that travis can safely run and expect passes from. I cannot think of any other good separation other than using the existing smoke/author/release categories that we have, but perhaps you might think of something more clever.

If the extra tests are too expensive to run constantly, then don't run them!  `dzil test --noa`, or even `prove -lr t`, are still available to you.

During development, I usually only use `prove` because even doing a build can take more time than I want to spend.

autarch:
@rjbs - if you feel strongly about it, I'd be up for redoing this to make the phase settable rather than just changing it unconditionally.

autarch:
I just rebased off master to see if that'll fix the test failures. @rjbs, any thoughts on this PR?

karenetheridge:
looks like the test failures now are due to JSON::PP, although it does look like JSON::PP is being upgraded to the latest version. I am mystified. I went over to the [@RJBS] queue to request the ability to add `Test::ReportPrereqs.include = JSON::PP` to dist.ini, so we can confirm that it really did get installed into the right place.

autarch:
nudge

rjbs:
Thanks, applied!

PR #474, opened 2015-07-15
PkgVersion can now use the first empty line it finds

dakkar:
Normally, `PkgVersion` inserts the version assignment just after the package declaration. This makes `Perl::Critic::Policy::TestingAndDebugging::RequireUseStrict` unhappy, because there's now code before `use strict`.

With this change, `PkgVersion` uses the first empty line it sees (before the next package declaration, if any). This not only avoids perlcritic problems, it also reduces the probability of line insertion, since most people will put a new line after all the `use`s.

### DIFF
--- a/lib/Dist/Zilla/Plugin/PkgVersion.pm
+++ b/lib/Dist/Zilla/Plugin/PkgVersion.pm
@@ -167,7 +167,7 @@ sub munge_perl {
   my %seen_pkg;
 
   my $munged = 0;
-  for my $stmt (@$package_stmts) {
+  while (my $stmt = shift @$package_stmts) {
     my $package = $stmt->namespace;
     if ($seen_pkg{ $package }++) {
       $self->log([ 'skipping package re-declaration for %s', $package ]);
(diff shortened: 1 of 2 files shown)


karenetheridge:
I fear this will break existing packages that do not have a blank line early in the package declaration.

FWIW, I do this to ensure that no code appears before strictures:

    use strict;
    use warnings;
    package Foo;

dakkar:
That incantation works very well, until you `use Moose` instead of an explicit `use strict`.

But you're right, presented with:
```
package Foo;
sub bar {

 return 1;
}
```
my code would produce:
```
package Foo;
sub bar {
$Foo::VERSION='1.00';
 return 1;
}
```
which is clearly very bad.
Let's see if I can make the code less wrong :)

karenetheridge:
> That incantation works very well, until you use Moose instead of an explicit use strict.

Why not both? :)

karenetheridge:
> Let's see if I can make the code less wrong :)

Good luck!  At the very least, you're going to learn a lot of PPI :)

rjbs:
Be sure to let me know when to review this. 😉

PR #476, opened 2015-07-19
drop the redundant -TRIAL in archive_filename if the version contains an underscore

### LINKED ISSUE #475: VERSION =~ /_/ forces -TRIAL in tarball name.
An annoying consequence of chained behaviour:

- [When version =~ /_/, release_status defaults to "testing"](https://metacpan.org/source/RJBS/Dist-Zilla-5.037/lib/Dist/Zilla/Dist/Builder.pm#L425)
- [When release_status is "testing", is_trial defaults to true](https://metacpan.org/source/RJBS/Dist-Zilla-5.037/lib/Dist/Zilla.pm#L550)
- [When is_trial is true, '-TRIAL' is always appeneded.](https://metacpan.org/source/RJBS/Dist-Zilla-5.037/lib/Dist/Zilla/Dist/Builder.pm#L425)

All this means the only way to force `is_trial` to be **false** when using a `_` version, you have to set the release status to a stable one ....

This is pretty much a "Go away" from Dzil.

karenetheridge:
resolves #475.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,8 @@
 Revision history for {{$dist->name}}
 
 {{$NEXT}}
+        - the distribution tarball name no longer includes -TRIAL if the
+          version contains an underscore
 
 5.039     2015-08-10 09:03:08-04:00 America/New_York
         - update required version of MooseX::Role::Parameterized; older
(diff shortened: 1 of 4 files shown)


karenetheridge:
This and #466 would constitute "make things work for underscore versions" and ought to go together in the same release IMO.

rjbs:
I feel a vague idealistic objection to this, but I think it's probably stupid, and at the very least I'm not about to do more work to show what I think would be better.  This solves a real complaint, so thanks, applied!

PR #480, opened 2015-07-27
(merge-ready) Always create a symlink to the latest build

karenetheridge:
This is incomplete, but I'm pushing it up now so I can ask questions.

What I'm going for is this: I want to create a `.build/latest` symlink no matter what type of build is run - even `dzil build`.

I've split the make-symlink logic out into its own private sub, and calling it in one extra place -- at the end of `build_in()` itself. However this is bad because now sometimes the symlink logic will be done twice, so the `previous` symlink will be lost.  I started moving this out and instead calling the symlink logic in the callers of `build_in()`, but that ends up duplicating information (i.e. the `.build` dir name). Alternatively, `build_in()` could be the one location where the symlink logic is called, but that requires build_in to have a return value that it currently doesn't have.

This could all be made a little cleaner if this bit of code was removed everywhere: `$latest->remove if $latest;` (as then we don't need to return the `$latest` symlink)...

thoughts? I'll keep trying to muddle ahead though...

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,7 @@
 Revision history for {{$dist->name}}
 
 {{$NEXT}}
+        - all builds now result in a .build/latest symlink
 
 6.012     2018-04-21 10:20:21+02:00 Europe/Oslo
         - revert addition of Archive::Tar::Wrapper as a mandatory prereq (in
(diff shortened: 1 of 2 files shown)


rjbs:
This looks pretty sane to me, I think.

rjbs:
I think this is waiting to be completed?

karenetheridge:
rebased to master (5.044) and I will review.

karenetheridge:
rebased to 6.005; this should be good to go now.

karenetheridge:
The "needs rework" tag can be removed now.

karenetheridge:
I think this is merged now?  I just noticed that I'd been rebasing a different branch than this one. so sorry you had to do that yourself.

PR #481, opened 2015-08-02
remove fatalized warnings

karenetheridge:
oops, there's more...


rjbs:
Thanks, merged.  What kind of monster could've written these commits to begin with?  Or merged them?

karenetheridge:
"If you realize you aren't so wise today as you thought you were yesterday, you're wiser today." - Olin Miller

But I'm going to go with "We're wiser today, so much so that we are appalled at the thought that we were once not so wise." - me

PR #482, opened 2015-08-05
omit unneeded content from Makefile.PL, Build.PL

karenetheridge:


rjbs:
Thanks, merged.  But I made a face while I did it!

PR #491, opened 2015-09-08
Use absolute paths rather than relative ones when gathering files

karenetheridge:
This fixes issues when using [ShareDir::Tarball] on OSX (where the sharedir is
inflated into /tmp and files are being gathered from it).

We can run into issues when the repo root is underneath a symlink at a
different depth than the real path -- e.g. OSX's TMPDIR defaults to /tmp,
which is a symlink to private/tmp (and /var symlinks to /private/var, so
/private/var/tmp is the final canonical location).

When we make /tmp/whargarbl/source/foo into a link relative to the repo root
(/tmp/whargarbl/source), we end up with ../../../private/tmp/whargarbl/source,
which does not exist.

Arguably this is actually an issue with File::Spec::abs2rel.

### DIFF
--- a/lib/Dist/Zilla/Plugin/GatherDir.pm
+++ b/lib/Dist/Zilla/Plugin/GatherDir.pm
@@ -173,7 +173,7 @@ sub gather_files {
   my $repo_root = $self->zilla->root;
   my $root = "" . $self->root;
   $root =~ s{^~([\\/])}{require File::HomeDir; File::HomeDir::->my_home . $1}e;
-  $root = path($repo_root)->child($root)->stringify if path($root)->is_relative;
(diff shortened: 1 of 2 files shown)


karenetheridge:
ha. There's the tests that I was sure I'd written that covered the same ground as we went over in #500. :)

cc @dagolden

rjbs:
Does this need a rebase, rework, or to be discarded?

karenetheridge:
I think what's here can still stand, and would replace the edit @dagolden made on that line.

karenetheridge:
This is still relevant. I've rebased it to the latest master, with the original branch preserved in https://github.com/karenetheridge/Dist-Zilla/tree/topic/gatherdir_symlinks_orig.

rjbs:
Thanks, applied!

PR #500, opened 2015-10-08
Fix GatherDir for minting with symlinked .dzil

dagolden:
Commit 387e3621edd3b1a1ff9aeef3e821866d68fc2eb1 breaks the search path when
the zilla root is a symlink.

Since the idea of that commit seemed to be to make the gather root relative
to zilla root when the gather root is relative, this commit does that more
directly using the 'child' method on the zilla root.

### DIFF
--- a/lib/Dist/Zilla/Plugin/GatherDir.pm
+++ b/lib/Dist/Zilla/Plugin/GatherDir.pm
@@ -173,7 +173,7 @@ sub gather_files {
   my $repo_root = $self->zilla->root;
   my $root = "" . $self->root;
   $root =~ s{^~([\\/])}{require File::HomeDir; File::HomeDir::->my_home . $1}e;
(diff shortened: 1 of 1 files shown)


karenetheridge:
If the problem could be reproduced with a test I would be quite grateful. I am struggling to see how the change makes a meaningful difference.

dagolden:
`X->child(Y)` means "tack Y onto X".  `Y->relative(X)` means "Y as a relative path assuming X is the cwd, resolving X or Y to real paths if they start relative".

That winds up with a root "../../etc..." relative to the repo path (which is actually a symlink), but File::File::Rule doesn't actually traverse the directories upwards, so it has to make root absolute, which is made absolute given the real path of cwd, not the the symlink repo path.  Since the symlink repo path and the real path have different numbers of directories, the "../../etc..." then points to the wrong place.

rjbs:
Thanks, applied.  Will release in short order.

Issue #487, opened 2015-08-10
Wishlist: ReportPhase functionality should be a part of Dist::Zilla.


van-de-bugger:
There is a nice plugin [`ReportPhase`](https://metacpan.org/pod/Dist::Zilla::Plugin::ReportPhase). However, there are few problems with it:

1. It seems it is not actively maintained now. There are few issues opened 10 months ago.
2. The plugin can be misused easily. For example, `[ReportPhase/Phase_Begins]` must be at the very top of `dist.ini`, while `[ReportPhase/Phase_Ends]` must be at the very bottom of `dist.ini`, otherwise user will have misleading output. (It is not clearly documented in `ReportPhase` documentation and is not obvious for a newbie developer.)
3. If `Dist::Zilla` introduces a new phase, the plugin should be updated.

`ReportPhase` functionality can be easily implemented directly in `Dist::Zilla`. For example,

    # Builder.pm, line ~348:
    $self->log_debug( 'gather_files begins...' ); # new line
    $_->gather_files       for @{ $self->plugins_with(-FileGatherer) };
    $self->log_debug( 'gather_files finished' ); # new line

It could be easily implemented as Dist::Zilla method to avoid repeating the same code in each phase:

    sub phase {
        my ( $self, $phase, $role ) = @_;
        $self->log_debug( [ '%s begins', $phase ] );
        $_->$phase for @{ $self->plugins_with("-$role") };
        $self->log_debug( [ '%s finished', $phase ] );
    };

then:

    # Builder.pm, line ~348:
    # $_->gather_files       for @{ $self->plugins_with(-FileGatherer) }; # old code
    $self->phase('gather_files', 'FileGatherer' ); # new code

Implementing such debug output directly in `Dist::Zilla` has numerous advantages:

1. No delay between adding a phase and supporting it by external plugin.
2. It cannot be misused, output is always correct.
3. Debug output can be turned on and off from command line, there is no need to add or remove lines in different parts of `dist.ini`.

rjbs:
Sounds reasonable.  I'd accept a PR.

van-de-bugger:
Does PR means "pull request"? 

`Dist-Zilla` repo has more than dozen of branches. Which branch should be forked?

rjbs:
Yes, and master.

rjbs:
I'm closing this because #501 replaces it.

PR #501, opened 2015-10-19
Wishlist: ReportPhase functionality should be a part of Dist::Zilla.

van-de-bugger:
Implementation of [Wishlist: ReportPhase functionality should be a part of Dist::Zilla.](https://github.com/rjbs/Dist-Zilla/issues/487)


rjbs:
Thanks, good start!

I'd like two changes before I merge this.

* the `phase` method should be private and called something like _call_plugin_method — I don't like the way that "phase" has crept into talking about the implementation
* the logged messages should be something like "now calling $method on all $role plugins"

These should be pretty simple.  Cool?

lib/Dist/Zilla.pm
```diff
@@ -799,6 +799,29 @@ sub stash_named {
   return $self->_global_stashes->{ $name };
 }
 
+=method _call_plugin_method
```
karenetheridge: Private methods aren't normally documented, because otherwise people will use them, and private methods are... private :)

lib/Dist/Zilla/Dist/Builder.pm
```diff
@@ -589,21 +588,23 @@ by the loaded plugins.
 sub release {
   my $self = shift;
 
-  Carp::croak("you can't release without any Releaser plugins")
-    unless my @releasers = @{ $self->plugins_with(-Releaser) };
+  $self->log_fatal("you can't release without any Releaser plugins")
```
karenetheridge: IMHO -- This should probably be pulled out into a separate change, because it doesn't really relate to this PR.

lib/Dist/Zilla/Dist/Builder.pm
```diff
@@ -776,13 +777,8 @@ does it clean up C<$directory> afterwards.
 sub run_tests_in {
   my ($self, $target, $arg) = @_;
 
-  Carp::croak("you can't test without any TestRunner plugins")
-    unless my @testers = @{ $self->plugins_with(-TestRunner) };
-
-  for my $tester (@testers) {
```
karenetheridge: Was it intended to remove these lines that call $tester->test for all testing plugins?

lib/Dist/Zilla/Dist/Builder.pm
```diff
@@ -776,13 +777,8 @@ does it clean up C<$directory> afterwards.
...
-    my $wd = File::pushd::pushd($target);
```
karenetheridge: This line is important and shouldn't be dropped.

lib/Dist/Zilla/Role/TestRunner.pm
```diff
@@ -45,4 +47,17 @@ around dump_config => sub {
   return $config;
 };
 
+=method test_in
+
+The method temporary switches to C<$target> directory and calls C<test> method,
```
karenetheridge: "temporary" -> "temporarily"
"calls C<test>" -> "calls the C<test>"

PR #504, opened 2015-11-01
Correct very minor grammar error in PkgVersion doc

paultcochrane:
Just happened to notice this tiny issue while reading the `PkgVersion` docs.

The conjugated verb needs to agree with the subject, in this case, "PkgVersion look" needs to be replaced by "PkgVersion looks".


rjbs:
Thanks, applied!

Issue #512, opened 2016-01-23
munge_files method allows code to screw up the iterated values


xsawyerx:
To summarize [this ticket](https://github.com/yanick/Dist-Zilla-Plugin-CoalescePod/issues/1) on [CoalescePod](https://metacpan.org/pod/Dist::Zilla::Plugin::CoalescePod):

The `munge_files` method (in `Dist/Zilla/Role/FileMunger.pm`) uses `$self->zilla->files` to iterate over the files. Then when it calls `munge_file` on each one, if a munger prunes a file from that attribute, the `munge_files` iteration will get screwed up.

Per @rjbs, "this isn't okay". However, this is also a problem with `CoalescePod` using the wrong phase. Hence, I would be happy if we had:

1. A fix to this in dzil (simple PR, but I want to make sure I'm not wrong).
2. A fix to CoalescePod - I'm not sure which phase should be used. Help?

Thanks!

karenetheridge:
I'll comment on the CoalesePod issue, as I don't think there is anything to do in Dist::Zilla itself.

PR #513, opened 2016-01-25
No longer iterate over a reference to zilla->files when munging files

mcmillhj:
PR for Issue #512 

[Issue #1](https://github.com/yanick/Dist-Zilla-Plugin-CoalescePod/issues/1) on Dist::Zilla::Plugin::CoalescePod brought to light an issue where if a FileMunger plugin were to issue a `$self->zilla->prune_file($file)` call this could, depending on the order of the files in `$self->zilla->files` mangle the file list, causing munge_file to not be called with a particular file. This PR corrects this by making a copy of `$self->zilla->files` before iterating.

A test, `t/plugins/munger-that-prunes.t`, has been included that reproduces this issue without the array copy mentioned above.

### DIFF
--- a/lib/Dist/Zilla/Role/FileMunger.pm
+++ b/lib/Dist/Zilla/Role/FileMunger.pm
@@ -29,7 +29,8 @@ sub munge_files {
   $self->log_fatal("no munge_file behavior implemented!")
     unless $self->can('munge_file');
 
-  for my $file (@{ $self->zilla->files }) {
+  my @files = @{ $self->zilla->files };
+  for my $file ( @files ) {
     if ($file->is_bytes) {
       $self->log_debug($file->name . " has 'bytes' encoding, skipping...");
       next;
--- a/t/lib/Dist/Zilla/Plugin/MungerThatPrunesPodFiles.pm
+++ b/t/lib/Dist/Zilla/Plugin/MungerThatPrunesPodFiles.pm
@@ -1,0 +1,17 @@
+package Dist::Zilla::Plugin::MungerThatPrunesPodFiles;
+
+use Moose;
+with(
+  'Dist::Zilla::Role::FileMunger',
+);
+
+sub munge_file {
+   my ( $self, $file ) = @_;
+   return unless $file->name =~ m/\.pod$/;
+   
+   $self->zilla->prune_file($file);
+}
+
+__PACKAGE__->meta->make_immutable;
+no Moose;
+1;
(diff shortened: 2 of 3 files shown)


karenetheridge:
I don't think this is a good change -- the list of files in the distribution should not be changing in the filemunging phase. After the pruning phase, the list of files should be considered locked down. (This is one of the things I test for in https://metacpan.org/pod/Dist::Zilla::Plugin::VerifyPhases.)

I presented a solution for CoalescePod in https://github.com/yanick/Dist-Zilla-Plugin-CoalescePod/issues/1#issuecomment-174635666.

mcmillhj:
While the issue from CoalescePod was the initial basis for this PR, I think there is still merit in passing a copy to the consuming classes in the event that they *do* use the phases improperly to get the result they desire. 

I think that your proposed solution makes sense, and can submit a PR to CoalescePod with those changes.

karenetheridge:
If a plugin uses the phases incorrectly, it's better to fail, IMO.

I'm not keen on receiving issues to `[Test::Compile]` (or any number of other plugins I ship that use the `:InstallModules` filefinder) complaining that it's referencing a non-existent file, and having to pick through the user's configuration only to discover that another plugin is to blame.

mcmillhj:
Fair enough.

rjbs:
I think you deleted this prematurely.  Sure, things should not abuse the ordering of actions, but this change is valuable not because it permits abuse, but because it prevents abuse from leading to really surprising action at a distance.  It would be better to avoid the weird behavior.

To prevent the abuse, the files should be controlled through methods like `add_file` and `remove_file`, and `files`.  Those methods can be restricted when they're not appropriate, like `remove_file` not working outside of a few actions.  One step toward doing that is to eliminate all use of the reference directly, and to use an elements accessor like `->files`.  That will provide semantics more like the ones in this PR.

The test, on the other hand, seems more like it is establishing more than just "no spooky action."  I'd be happy to apply the change to avoid the problem, although I'm less thrilled about the test as written.  Either let's skip it or rework it to present the situation as anomalous and better avoided not something to be guaranteed.

I can't re-open this PR, because the branch has been deleted.

mcmillhj:
My mistake, I am happy to rework the test and patch based on any advice you have.

rjbs:
Thanks, I will try to merge this in the near future.  Digging out atm.…

mcmillhj:
Oh, I was going to rework the test before this was merged. You mentioned that you didn't like it in it's current state. Any advice on how to make it better? I have used Dist::Zilla for awhile but blissfully ignored the internals, so I put this together by looking at other FileMunger tests.

rjbs:
I've just added a comment for myself in the test and applied the patch.  Thanks!

PR #515, opened 2016-02-12
Make FakeRelease compatible with UploadToCPAN

Csson:
Hi!

This change makes [FakeRelease] similar enough to [UploadToCPAN] to be usable in testing of [NextRelease] type plugins.


karenetheridge:
I guess `[FakeRelease]` could also be modified directly in the test (similarly to how `[UploadToCPAN]` was modified to neuter the `release` sub), rather than in the source.  But I can also see the advantage of being able to use `[FakeRelease]` with a plugin bundle that uses `[NextRelease]` and `%P`.

rjbs:
Thanks, applied as provided!

PR #517, opened 2016-02-12
Remove duplicates from the :ExecFiles result

karenetheridge:
This can happen if there are multiple ExecFiles plugins providing overlapping
results.


rjbs:
Thanks, applied!

PR #522, opened 2016-02-24
Use a String::RewritePrefix rewriter

dolmen:
I noticed that `Dist::Zilla::Util::expand_config_package_name` is called 56 times during a `dzil nop` on the dzil repo. And that it doesn't use a [String::RewritePrefix](https://metacpan.org/pod/String::RewritePrefix) rewriter.
So here is the rewrite of this sub using a rewriter built once.

See also my improvements on String::RewritePrefix itself: rjbs/String-RewritePrefix#2


karenetheridge:
:+1:

rjbs:
Great idea, thanks!

PR #523, opened 2016-03-13
replace uses of List::MoreUtils with List::Util

karenetheridge:
Rewrite of #456; this should be much nicer!


rjbs:
Thanks, applied.  Definitely better.

karenetheridge:
I'm following up with a little tweak to this. don't freak out! :)
(ps don't forget to push your master branch)

PR #524, opened 2016-03-13
support different prereq relationships in [AutoPrereqs]

karenetheridge:
It was pretty easy to add, and some dist.inis can be simplified as a result (e.g. MooseX-Runnable, which makes heavy use of "suggests" prereqs via its optional plugin moduels).

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,7 @@
 Revision history for {{$dist->name}}
 
 {{$NEXT}}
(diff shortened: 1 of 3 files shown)


rjbs:
Okay, why not?

Applied to my local repo, thanks!  Special thanks for the Changes update.

dolmen:
@karenetheridge As I'm working with @book about refactoring AutoPrereqs, we would like to discuss about your idea...

PR #525, opened 2016-04-01
fix class_type import

karenetheridge:
using MooseX::Types directly puts MooseX::Types::Base into @ISA!

(as spotted by @kentnl on #distzilla, 2016-03-31 18:20 PDT)


rjbs:
Woah!

Thanks, applied to my local repo!

PR #526, opened 2016-04-01
fix documentation for found_files method

karenetheridge:
This sub returns a listref, not a list.


rjbs:
Thanks, applied in my local repo!

PR #527, opened 2016-04-01
reject version ranges rather than warning, if eumm_version is not high enough

karenetheridge:
This mirrors changes recently released in `[MakeMaker::Awesome]` version 0.37.


rjbs:
Thanks, applied!

As per IRC discussion, I'd like it even better if it would:

* automatically bump up EUMM version required if none had been set by user
* otherwise die

PR #529, opened 2016-04-07
parentheses must be used for subs that were not loaded at compile time

karenetheridge:
This may break eventually without this change, just as using LMU just broke when Moose stopped using it.


dolmen:
Or just load `List::Util` with a `use` statement at compile time. I don't think that using lazy loading for List::Util is necessary. Especially if Moose already loads it.

rjbs:
Thanks, applied in my local repo!

PR #535, opened 2016-04-21
Make 'run' respect the --trial option

karenetheridge:
The 'run' command needs to respect the same options as 'build', so it can
replicate the same build environment (e.g. plugin configs and build outputs).


rjbs:
Thanks, applied!

PR #536, opened 2016-04-21
Improve DZP::FileFinder::ByName::_join_re

dolmen:
Faster `_join_re()`:
- do not recompile regexps already compiled by callers (either by type coercion or by explicit `qr//`)
- special case a single element list to avoid regexp stringify+compile
- just `join('|')` to combine multiple regexps
- compile the final result with `qr//`


dolmen:
Rebased.

rjbs:
thanks, applied!

PR #539, opened 2016-04-24
change directories first before absolutising @INC

karenetheridge:
and, a test!


rjbs:
Thanks!  I applied this, but with a tweak in e795c1b.

PR #545, opened 2016-05-02
strip Path::Class from prereqs

karenetheridge:
Avoid specifying Path::Class as a prereq, since it is only used in fallback code in Dist::Zilla::Path where some other plugin would already have a Path::Class dependency.

(should get at least a week in -trial to shake out any new downstream issues.)


rjbs:
Thanks, applied.

PR #548, opened 2016-05-15
Fix some problems when skipping tests under Win32

klaus03:
This fixes some test failures under Win32


klaus03:
I just realized that the original TODO should really be a SKIP, hence my second commit...

dolmen:
It seems to be a duplicate of #540 and it has been fixed in the `dev-v6` branch.

karenetheridge:
dev-v6 is now back in master so this should all be good now.

rjbs:
Closing this on the theory that it's fixed by #540

PR #549, opened 2016-05-16
new plugin to allow gathering individual files

karenetheridge:


lib/Dist/Zilla/Plugin/GatherFile.pm
```diff
@@ -0,0 +1,135 @@
...
+
+sub gather_files {
+  my ($self) = @_;
+
+  my $repo_root = $self->zilla->root;
+  my $root = "" . $self->root;
+  $root =~ s{^~([\\/])}{require File::HomeDir; File::HomeDir::->my_home . $1}e;
+  $root = path($root);
+  $root = $root->absolute($repo_root) if path($root)->is_relative;
+
```
dolmen: I think that gathering files outside the project directory (either with an absolute path or relative to $HOME) is quite insane.

karenetheridge: Perhaps, but GatherDir supports this, and it is occasionally useful. :)

rjbs:
Thanks, applied!

Issue #546, opened 2016-05-08
DZP::ExtraTests should write dependency-free skip logic


kentfredric:
As seen in https://rt.cpan.org/Ticket/Display.html?id=114237 , excessive application of DZP::ExtraTests wastes a lot of CPU time to basically print stuff to the screen.

Its annoying enough to waste CPU time to display it, but its even more annoying having such a large volume of "use Test::More" simply to print a "skipped" line and exit.

Seems this is a very good candidate for not loading Test::More.

dolmen:
ExtraTests is an awful solution to a legitimate issue. We have now DZP::RunExtraTests (but not in dzil core) that does the job much better.
I suggested to @RJBS to deprecate ExtraTests in dzil v6.

kentfredric:
It is an awful solution, indeed. But we can mitigate the overhead of using it slightly for those who insist on using it.

(Fwiw, I habitually nuke re-written tests in the vendor side testing chain because their existence just causes confusion and doesn't really help anyone)

kentfredric:
Probably worth mentioning the tax of this approach is greatly amplified by Test2's increase of modules.

Its probably the worst-case for Test2.

dolmen:
But Test2 also allows to change the formatter to something else than TAP which this would not allow:
```
BEGIN { unless ($ENV{RELEASE_TESTING}) { print "1..0 # SKIP these test are for release candidate testing\n"; exit } }
```

kentfredric:
> But Test2 also allows to change the formatter to something else than TAP which this would not allow:

Quite, but given the proliferation of competing ways to emit tap that are used in the wild ( one of which is legacy Test.pm ), worrying "It could break because some harness might stop supporting TAP" seems like the worry is in the wrong place.

I'd be much more worried if Test::Harness decided it was a good time to start rejecting test-output emitters other than Test::More/Test2 

But then again...... I'm not sure what to expect anymore.

dolmen:
I was only raising the Test2 formatter feature because the use case you linked in the initial report is a project from DROLSKY and [he clearly expressed his interest in using alternate formatters thanks to Test2](http://blogs.perl.org/users/chad_exodist_granum/2016/04/test2testbuilder-update-from-the-qah.html#comment-1702928).

kentfredric:
Ah. Right. Fair point. 

So, in your estimation, would it make sense to support an option that dictated it would emit a "thin" rewrite instead of a "fat" rewrite?

[...]

PR #551, opened 2016-05-17
#546 ExtraTests: print TAP directly for faster skip

dolmen:
Fix for #546: faster skip_all by printing TAP ourself instead of loading Test::More for that.


rjbs:
Thanks, merged with removal of unneeded single quotes.

PR #552, opened 2016-05-18
do not chdir into the source dir when building

karenetheridge:
I read "we must!" and "but it doesn't work!" and I thought "OH REALLY, what doesn't work?" and it turns out that there are only two things in core that don't prepend root properly. There may be plugins out there that are naughty. Let's find out!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ perl:
   - "5.16"
   - "5.14"
 before_install:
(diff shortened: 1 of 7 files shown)


rjbs:
I want to release this soon.  I want to put this in v7.  Therefore I will try to do v7 in less than three years.

karenetheridge:
rebased to new master, which clears up the junk from the dev/v6 merge that was appearing before.

karenetheridge:
rebased to 6.009.

rjbs:
Thanks, applied locally!

PR #553, opened 2016-05-18
treat missing Changes as a fatal error

karenetheridge:
as discovered by abraxxa on #distzilla, 2016-05-18


rjbs:
Thanks, applied!

PR #554, opened 2016-05-18
add [MetaJSON] to [@Basic]

karenetheridge:


dolmen:
It would be helpful to document this v6 change in the Pod of `[@Basic]`.

lib/Dist/Zilla/PluginBundle/Basic.pm
```diff
@@ -45,6 +46,7 @@ It includes the following plugins with their default configuration:
 * L<Dist::Zilla::Plugin::PruneCruft>
 * L<Dist::Zilla::Plugin::ManifestSkip>
 * L<Dist::Zilla::Plugin::MetaYAML>
+* L<Dist::Zilla::Plugin::MetaJSON>
```
karenetheridge: @dolmen The plugin membership is listed here -- where else should it go?

dolmen: I was think about something more visible (well, as far as POD for a PluginBundle can be considered visible) such as a block in the DESCRIPTION section:
```pod
B<Note:> L<[MetaCPAN]|Dist::Zilla::Plugin::MetaJSON> is included since Dist::Zilla v6. 
```

rjbs:
I'm not sure how I feel about special casing "can appear twice, whatever!" in a specific plugin, rather than just forcing people to update.  I'm going to put this off for now.

karenetheridge:
The problem with having two MetaJSONs being a fatal error is that if then the distribution is built against Dist::Zilla 5, now they have no META.json file at all, which I think is a worse outcome for CPAN.

karenetheridge:
note to self: need to hack the error message to say something like "to fix this, remove \[MetaJSON\] from your dist.ini, and add :version = 7.000 underneath \[\@Basic\]"

karenetheridge:
updated to v7-in-progress.

Issue #544, opened 2016-05-02
"dzil listdeps --missing" fails to detect Sub::Util


dolmen:
If the distribution has a dependency on Sub::Util > 0 (e.g. Sub::Util 1.40), `dzil listdeps --missing` still reports Sub::Util must be installed even if Sub::Util 1.45 (latest from CPAN) is installed.

This problem is due to an issue with `Class::Load::{PP,XS}::is_class_loaded`. See [RT#114126](https://rt.cpan.org/Ticket/Display.html?id=114126), [RT#114125](https://rt.cpan.org/Ticket/Display.html?id=114125).

The fix: Dist::Zilla::App::Command::listdeps and Dist::Zilla::Util::AutoPrereqs must use Module::Runtime instead of Class::Load to detect modules because Module::Runtime uses the presence in %INC instead of the presence of a stash.

dolmen:
Workaround: `perl -MSub::Util -S dzil listdeps --missing`

karenetheridge:
I happily endorse the idea of switching to Module::Runtime.

rjbs:
Fine with me.

dolmen:
I have work in progress on this issue...

rjbs:
I guess this is fixed by merging #555?

PR #555, opened 2016-05-19
#544 listdeps/authordeps --missing with Module::Runtime

dolmen:
Fix for #544: `dzil listdeps --missing` and `dzil authordeps --missing` reimplemented using Module::Runtime instead of Class::Load.


rjbs:
Thanks, applied!

PR #556, opened 2016-05-25
more x_breaks data

karenetheridge:
If @xdg releases `[CheckChangesHasContent]` before you get to this, then this line should also be added:

    Dist::Zilla::Plugin::CheckChangesHasContent = < 0.007


xdg:
FYI: I have just released an updated CheckChangesHasContent.

karenetheridge:
hmm, actually, CheckChangesHasContent shouldn't be listed... only its tests were broken, not the actual code.  If the plugin is installed first and then Dist::Zilla upgraded on top, things should still work fine.

rjbs:
thanks, applied

PR #558, opened 2016-06-12
Populate the main section of metadata after calling MetaProviders

karenetheridge:
This allows MetaProvider plugins to make modifications to read-write
attributes on $zilla before they are committed to metadata, e.g. abstract.


rjbs:
thanks, applied

PR #559, opened 2016-07-05
restrict [MetaYAML] to metaspec v1.4, [MetaJSON] to v2.0+

karenetheridge:


rjbs:
I don't know that I care about people being able to footgun, but you have convinced me by making it an overall drop in lines of code. :wink:  Applied.

Issue #497, opened 2015-10-03
Tests fail under Cygwin Perl 5.22


vadz:
Here is the end of `cpanm` log:
```
PERL_DL_NONLAZY=1 "/usr/bin/perl.exe" "-MExtUtils::Command::MM" "-MTest::Harness" "-e" "undef *Test::Harness::Switches; test_harness(0, 'blib/lib', 'blib/arch')" t/*.t t/commands/*.t t/diagnostics/*.t t/plugins/*.t
# 
# Versions for all modules listed in MYMETA.json (including optional ones):
# 
# === Configure Requires ===
# 
#     Module                  Want    Have
#     ----------------------- ---- -------
#     ExtUtils::MakeMaker      any 7.04_01
#     File::ShareDir::Install 0.06    0.10
# 
# === Build Requires ===
# 
#     Module              Want    Have
#     ------------------- ---- -------
#     ExtUtils::MakeMaker  any 7.04_01
# 
# === Test Requires ===
# 
#     Module                    Want     Have
#     ------------------------ ----- --------
#     CPAN::Meta::Check        0.007    0.012
#     CPAN::Meta::Requirements 2.121    2.132
#     ExtUtils::MakeMaker        any  7.04_01
#     ExtUtils::Manifest        1.66     1.70
#     File::Spec                 any     3.56
#     Software::License::None    any 0.103010
#     Test::FailWarnings         any    0.008
#     Test::Fatal                any    0.014
#     Test::File::ShareDir       any 1.001001
#     Test::More                0.96 1.001014
#     lib                        any     0.63
#     utf8                       any     1.17
# 
# === Test Recommends ===
# 
#     Module         Want     Have
#     ---------- -------- --------
#     CPAN::Meta 2.120900 2.150001
# 
# === Runtime Requires ===
# 
#     Module                                         Want       Have
#     ------------------------------------------ -------- ----------
#     App::Cmd::Command::version                      any      0.328
#     App::Cmd::Setup                               0.309      0.328
#     App::Cmd::Tester                              0.306      0.328
#     App::Cmd::Tester::CaptureExternal               any      0.328
#     Archive::Tar                                    any       2.04
#     CPAN::Meta::Converter                      2.101550   2.150001
#     CPAN::Meta::Merge                               any   2.150001
#     CPAN::Meta::Prereqs                        2.120630   2.150001
#     CPAN::Meta::Requirements                      2.121      2.132
[...]

PR #563, opened 2016-08-12
Don't depend directly on DateTime::TimeZone::Local::Win32

schwern:
As of 1.92, DateTime::TimeZone will take care of this. Depending
on it directly risks DT::TZ and DT::TZ::Local::Win32 being out
of sync.

See https://rt.cpan.org/Ticket/Display.html?id=116909 and https://ci.appveyor.com/project/schwern/dist-zilla/build/1.0.4

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -45,9 +45,6 @@ Mixin::Linewise::Readers = 0.100    ; default encodings to UTF-8
 
 DateTime = 0.44 ; CLDR fixes, used by AutoVersion and NextRelease
 
-[OSPrereqs / MSWin32]
-DateTime::TimeZone::Local::Win32 = 1.80
-
 [RemovePrereqs]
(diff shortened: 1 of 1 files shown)


karenetheridge:
The DateTime::TimeZone::Local::Win32 dependency was added to fix https://github.com/rjbs/Dist-Zilla/issues/497.

schwern:
Maybe depend on DateTime::TimeZone 1.92 in Win32 instead? That will pull in DateTime::TimeZone::Local::Win32 and won't fall out of sync with DateTime::TimeZone.

Maybe @autarch has an insight?

autarch:
We don't want to have circular deps. The thinking was that it's best to have the dep in DT::TZ since that's the thing we're sure Windows users will install. Otherwise they'd have to manually install this distro too (and how would they even know it exists?).

PR #564, opened 2016-08-14
Depend on DateTime::TimeZone directly on Win32.

schwern:
DT::TZ::Local::Win32 is a Win32 specific component of DT::TZ and
should not be dependend on directly else it might be out of sync
with an installed version of DT::TZ.

DT::TZ 1.92 is when it started depending on DT::TZ::Local::Win32.

See #563 for more discussion.


rjbs:
Given the closure of #563, is this also to-close?

schwern:
No, AFAIK the problem still exists. This is an improved version of #563 based on [@autarch's comment](https://github.com/rjbs/Dist-Zilla/pull/563#issuecomment-239625496).

rjbs:
Okeydoke, applied.

PR #562, opened 2016-08-12
Add a config for AppVeyor.

schwern:
It's CI for Windows! It'll test against Strawberry Perl.

Go to https://appveyor.com to sign up and turn it on.

<strike>This will currently fail because of #563.</strike>
That got fixed, now it's failing for other reasons.

PR #565, opened 2016-08-14
Add listdeps options for requires/recommends/suggests.

schwern:
This allows more control over dependency listing.

`dzil listdeps` includes the recommended deps. The recommended Dist::Zilla deps Term::ReadLine::Gnu and Archive::Tar::Wrapper are unlikely to work on Windows. Allowing `dzil listdeps --no-recommends` will allow #562 to work and make it easier for Windows users to work with dzil from the repo.

### DIFF
--- a/corpus/dist/AutoPrereqs/dist.ini
+++ b/corpus/dist/AutoPrereqs/dist.ini
@@ -18,3 +18,10 @@ filename = t/data.bin
 
 [Prereqs / DevelopRequires]
 String::Formatter = 0
+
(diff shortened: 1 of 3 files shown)


rjbs:
When applied to master, this produces failures in t/commands/listdeps now, but I think they're exposing a real bug in cpanm-style version output.  I'll dig…

rjbs:
Thanks, applied with some further changes to tests to improve coverage of combinations of listdeps switches!

PR #566, opened 2016-09-07
fix ExtraTests skip message quoting

rivy:
This is needed since the smoke testing message contains internal double-quotes (i.e., '"smoke bot" ...').


dolmen:
This patch doesn't fix anything. There is no issue in the original code.

karenetheridge:
@dolmen normally that's true, but `$msg` in this code is filled in at build time, not when the generated test is executed.  Without the patch, the generated line looks like:

    print "1..0 # SKIP these tests are for "smoke bot" testing\n";

dolmen:
Sorry @rivy. I'm wrong. When reviewing on GitHub I missed that the whole BEGIN block code was itself in a string.

rjbs:
Thanks, applied!

PR #571, opened 2016-10-19
Add cpanm versions test

mickeyn:
noticed this is not tested + conditions seem a bit too complicated than needed.


rjbs:
Thanks, applied!

PR #573, opened 2016-12-17
also run Module::Runtime::Conflicts in t/zzz-check-breaks.t

karenetheridge:
This doesn't add any additional prereqs, because Module::Runtime::Conflicts is
a prereq of Moose.


rjbs:
Applied, thanks!

PR #575, opened 2016-12-17
also prune tmp/

karenetheridge:
Dist::Zilla::Tester might leave this (non-empty) directory behind if some
tests failed, but it should never be included in the distribution.


rjbs:
I am 95% on board and 5% worried about somebody who is using GatherDir::Git with PruneCruft.  But presumably you should never need the latter and former together.  Applied!

PR #576, opened 2016-12-17
treat Module::Runtime as an "internal" module for carping

karenetheridge:
This means that if a module (e.g. a Dist::Zilla plugin) being loaded via
Module::Runtime issues a warning, we will look to the caller for warning
categories, rather than Module::Runtime itself (which happens to disable *all*
warnings).


rjbs:
Not fully convinced.  Convinced enough.  Merged!

PR #577, opened 2017-01-03
detect :version directive in root section in authordeps

karenetheridge:


dolmen:
I understand the rationale for calling this key `:version` (for consistency with `:version` in plugins sections) but I think this will only bring confusion to users who will have both `version` and `:version` in the top section.
If the key need a comment just above the line to say "; this is the required dzil version" because it is obscure, I think it would be just better to call the key `dzil_version`.

karenetheridge:
This is a very rarely-needed feature; it's not even documented.  Generally
it's the plugins that have a specific version requirement, not core itself.

rjbs:
Thanks, good idea, and applied!

PR #578, opened 2017-01-08
Format copyright year as "X - Y" when the year in the config is not t…

autarch:
…he current year

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,9 @@
 Revision history for {{$dist->name}}
 
 {{$NEXT}}
+        - if the copyright_year in the dist.ini is not this year, the
+          copyright statement and license are formatted as "$year -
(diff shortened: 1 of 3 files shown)


autarch:
Somewhat related to #135

rjbs:
I really don't care for this kind of change, because opinions and lawyers seem to vary on what is appropriate under what circumstances.  Would a CopyrightProvider be better?  I don't know, seems like a pain.  I'll think about it.

autarch:
Yeah, all I really want is a way to format the year a certain way without having to provide my own License class.

rjbs:
If I had a time machine, I'd have used `copyright_since` or something.

rjbs:
addressed in #745

PR #582, opened 2017-03-22
Documentation fix

regina-verbae:
Fix documentation to use proper method in document_assigns_to_variable example.


rjbs:
Thanks, applied!

Issue #521, opened 2016-02-17
`dzil build` should empty the target directory when starting up


karenetheridge:
as noted here -- https://rt.cpan.org/Ticket/Display.html?id=111984#txn-1596865 -- if the build directory already exists, `dzil build` will add its files to the existing contents rather than emptying it out. This can yield undesired results if the build configuration was changed, as files can be left over from last time that didn't get overwritten.  The directory should probably be emptied out at the start of the build.

rjbs:
…and probably prompt before doing so.  Sounds good.

tlhackque:
I don't object to a prompt - but there should be a way to suppress it.

Usually the build directories are a scratch area that can be iterated on many times with trial runs & packaging experiments.  And people shouldn't be doing work inside them.  (though some probably do.)

So while a prompt isn't a bad thing - it could, for example, catch a case where one forgets to update a version number - it should be possible to avoid the prompt.

Perhaps a dist.ini options such as PromptIfBuildDirectoryDirty = yes|no.

I won't quibble about the default.

rjbs:
That all sounds fines to me except for that option name.  Good grief, man! 😄

karenetheridge:
It could be a command-line option on the `build` command? (and `test`, `release`, `smoke` too, I guess...), to avoid promping? I guess I don't mind which way it defaults as I use a shell alias to do a build anyway.

karenetheridge:
For the `release` command especially I think it's important to empty out the directory first -- otherwise the uploaded distribution isn't exactly what was intended to be built.

tlhackque:
Besides shipping what you intend, you'd really like your tests prior to release to run in the same environment as release.

To be clear, if you say 'no' to the prompt, dz should abort the build.

A build should **always** empty the directory.  The only point to a prompt is to allow for _oops, for some reason the wrong directory is being used._  E.g. the build directory name comes from the version, which sometimes people forget to update.  

A dz build proves that your kit is complete.  If it relies on left-over junk, it's a false proof.  If it ships stuff that found its way into the build directory by some other path, the shipped kit is incorrect.  Maybe mildly so.  Maybe some confidential material gets packaged...

**Any build in a non-empty directory is defective**.  The only choices are to change to another directory, or empty it.

rjbs:
Yes, agreed: the point of the prompt is not to allow you to not empty the dir and still use.  It's to allow you to abort and fix the dir you're using, if needed.

[...]

PR #583, opened 2017-04-04
Add more info to the help for the build command

genio:
Just goes a bit further in depth on explaining with the ```dzil build``` command will do.


karenetheridge:
I would clarify that the making of the build dir is not a default action, but something that the build action *always* does; by contrast, creating the tarball is controlled by the 'tar' option (which defaults to on, and is disableable via --notar.

genio:
That's a good point. I've reworded it a bit to hopefully clarify that. If you can read over it again, I'd appreciate it.

lib/Dist/Zilla/App/Command/build.pm
```diff
@@ -16,6 +16,38 @@ does all the things required to build your distribution.  By default, it will
...
+
+=over
+
+=item
+
+Generate a directory containing your module, C<Foo-0.100>. This directory is
+complete. You could create a gzipped tarball from this directory and upload it
+directly to C<PAUSE> if you so desired. You could C<cd> into this directory and
+test your module so you're certain everything is as it should be before
+uploading to C<PAUSE>.
```
karenetheridge: I don't think we should suggest doing this, as it can cause files such as MYMETA.json to be left behind in the tarball, which should never appear in an uploaded distribution.  (And presently, even doing another 'build' afterward won't actually wipe out what was there before -- see issue #521 -- one would have to explicitly delete the directory, or run `dzil clean`).  `dzil test` is much safer for testing the distribution.

genio: Ah, I wasn't trying to lead people down that path at all, but rereading it I can certainly see how it could be taken that way.  I personally test this way on multiple perls with ```perlbrew exec ...``` but I don't upload to PAUSE from the directory. I upload from the gzipped tarball that's created at the same time.
My goal was to get across the point that the freshly generated directory is _exactly_ what you would get from the tarball or what would be uploaded if you did a ```dzil release```.  I'll mull over how to word that a bit more clearly.

genio:
I _think_ it reads better now.

rjbs:
Thanks, applied!

PR #590, opened 2017-05-08
continue to find [=inc::SomePlugin] even when . is not in @INC

karenetheridge:
I think it's probably safe now to remove the chdir into dist_root during test builds, but this change should be done separately (and in a -TRIAL release) in case some other distributions' tests were depending on it.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,8 @@
 Revision history for {{$dist->name}}
 
 {{$NEXT}}
+        - continue to support local plugins like [=inc::SomePlugin] when . is
+          no longer in @INC
 
 6.009     2017-03-04 11:16:37-05:00 America/New_York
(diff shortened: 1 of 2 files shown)


karenetheridge:
now in the light of day I need to revise this. it only fixes the tester, not plugins themselves. One of the core tests does break in blead without this change, but is it the right one? should we make users doctor their own lib (e.g. see the `[lib]` plugin), or should we fix it for them?

karenetheridge:
This needs to be fixed soon - as currently Dist-Zilla requires PERL_USE_UNSAFE_INC=1 to be set for installation on perl 5.26+ (and 5.22, 5.24 on debian)

rjbs:
I believe this is now fixed by https://github.com/rjbs/Config-MVP/commit/3966df4193d20ac8aaa639d2a6b65875431eba3f

PR #593, opened 2017-05-23
[NameFromDirectory] 0.03 and earlier assume $zilla->root is a Path::C…

karenetheridge:
…lass::Dir


karenetheridge:
this is an easy patch - please merge when you can

rjbs:
Thanks, applied!

PR #595, opened 2017-07-02
add a debug log line when a Dist::Zilla::File::FromCode sub is run

karenetheridge:


rjbs:
Thanks, applied!

rjbs:
…and now all the tests fail, because Files don't have log_debug delegated.

…which means you probably didn't run the tests before making this PR?

karenetheridge:
doh!!!

is there any good way of accessing the logging object from a file?

rjbs:
I don't think so.

karenetheridge:
reluctantly withdrawing this PR.  a simple `warn` would be ok, if there was a global DEBUG variable or constant to tap into, but I don't think one exists.

PR #596, opened 2017-07-03
Added some examples to the documentation

jonasbn:
Added SYNOPSIS section and example with filename parameter.


rjbs:
Thanks, applied!

PR #597, opened 2017-07-24
find_files comes from Dist::Zilla::Role::ExecFiles

wolfsage:


karenetheridge:
lol, all the time this role was useless because it was being overridden locally with a copy of the same sub.

it's been like this forever since these two packages were written back in 2010-03-22.

karenetheridge:
maybe this change should be the other way, with the `find_files` sub disappearing in the role and replaced with `requires 'find_files';` -- the role is being used as a marker for the `:ExecFiles` file finder to find the appropriate plugin(s) to interrogate. all that matters is that the composing classes implement `find_files`.

rjbs:
For now, the given fix makes the most sense to me.  Applied!

PR #598, opened 2017-08-13
replace uses of Class::Load with Module::Runtime

karenetheridge:


karenetheridge:
Class::Load is mostly just a wrapper around Module::Runtime, but adds a few weird heuristics that are best avoided unless absolutely necessary (which they aren't, here).

rjbs:
Thanks,applied!

Issue #581, opened 2017-02-27
t/tester-local-plugins.t fails test without '.' in @INC


kentfredric:
Unsure if this is related to #580 or not

```
t/zzz-check-breaks.t ............ ok

#   Failed test 'config does not blow up with "Required plugin inc::MyMetadata isn't installed."'
#   at t/tester-local-plugins.t line 36.
#          got: 'Required plugin inc::MyMetadata isn't installed.
# 
# Run 'dzil authordeps' to see a list of all required plugins.
# You can pipe the list to your CPAN client to install or update them:
# 
#     dzil authordeps --missing | cpanm
# 
# '
#     expected: undef
# Looks like you failed 1 test of 1.
t/tester-local-plugins.t ........ 
Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/1 subtests 
```

latk:
Aka. “Dist::Zilla 6.010 can't load plugins from `inc/` under perl 5.26”. As a workaround one can `export PERL_USE_UNSAFE_INC=1` in the current shell, but this is really something that dzil should support by itself.

Note that in some cases dzil handles `inc::` modules specially, e.g. by prioritizing them in the `authordeps --missing` command (see the [`Dist::Zilla::Util::AuthorDeps::extract_author_deps()` function. This prioritization seems to make it impossible to consistently work around the `inc/` handling via the (external) [`lib` plugin](https://metacpan.org/pod/Dist::Zilla::Plugin::lib), but I haven't looked into that deeply.

I have no idea how this should be resolved, possibly by deprecating `inc::` style plugins (please not) or by adding the root dir to `@INC` while loading plugins (also seems ugly), or by supporting built-in functionality like the `lib` plugin, but special-casing it where necessary.

kentfredric:
> Aka. “Dist::Zilla 6.010 can't load plugins from inc/ under perl 5.26”. As a workaround one can export PERL_USE_UNSAFE_INC=1 in the current shell, but this is really something that dzil should support by itself.

Yeah. There turns out to be no really sane way of handling this other than ensuring "." is reinstated globally sometime before loading the first plugin. Given that by design, you *MUST* be in your project root for `dzil` to even work, the typical paranoia about ". might be insecure" is moot.

[...]

PR #600, opened 2017-10-06
Keep dist-root in @INC beyond Perl 5.26

kentfredric:
The only thing that makes sense is to ensure CWD of some description is
in @INC at some stage before plugins are actually loaded, because
that's the traditional expectation.

Basically, this option is the only option that works. There's no viable
way to scope @INC changes to anything narrower without breaking the
usage entirely.

There's no sane way to limit the depth of @INC changes either, even if
we wanted to.

And if you have vulnerable code in your project source tree that you're
worried about, you're pretty much beyond screwed already, and you're
about to be screwing CPAN when you ship next, and no amount of @INC
magic will save you.

Closes: https://github.com/rjbs/Dist-Zilla/issues/581
Closes: https://github.com/rjbs/Dist-Zilla/pull/590
Bug: https://bugs.gentoo.org/613586

### DIFF
--- a/lib/Dist/Zilla/Dist/Builder.pm
+++ b/lib/Dist/Zilla/Dist/Builder.pm
@@ -36,6 +36,7 @@ sub from_config {
   $arg ||= {};
 
   my $root = path($arg->{dist_root} || '.');
+  unshift @INC, $root->absolute->stringify;
 
   my $sequence = $class->_load_config({
     root   => $root,
--- a/lib/Dist/Zilla/Util/AuthorDeps.pm
+++ b/lib/Dist/Zilla/Util/AuthorDeps.pm
@@ -105,7 +105,7 @@ sub extract_author_deps {
 
   if ($missing) {
     require Module::Runtime;
-
+    unshift @INC, $root->absolute->stringify;
     @packages =
       grep {
         $_ eq 'perl'


karenetheridge:
> Basically, this option is the only option that works. There's no viable
way to scope @inc changes to anything narrower without breaking the
usage entirely.

How so? \@INC should only be modified for plugin loading, and no other time. It's possible to limit the scope just to that place.

kentfredric:
> How so? @INC should only be modified for plugin loading, and no other time. It's possible to limit the scope just to that place.

Not without breaking the ability for plugins to extend `@INC`, which would break .... `[lib]` and everything like it.

karenetheridge:
I'm not terribly concerned about breaking `[lib]` if the reason for its existence is fixed in core dzil.  Is there ever else a need to modify `@INC` in a non-localized way?

kentfredric:
[Bootstrap::ShareDir ... ]

kentfredric:
And also, [Bootstrap::lib], works that, way, but has the option of using a previously generated library in a previously generated copy of your own dist in a previous ./Dist-Name-Here/ directory and other tricks like that.  `[lib]` is only the lightweight version for people who don't need or want those kinds of hoops.

rjbs:
https://github.com/rjbs/Config-MVP/commit/3966df4193d20ac8aaa639d2a6b65875431eba3f

PR #603, opened 2017-12-10
remove extraneous whitespace from generated README files

karenetheridge:


rjbs:
I applied 599, which accomplishes the same goal.  Thanks!

PR #604, opened 2017-12-10
remove all uses of File::HomeDir

karenetheridge:
...as inspired by File::HomeDir::Tiny

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,7 @@
 Revision history for {{$dist->name}}
 
 {{$NEXT}}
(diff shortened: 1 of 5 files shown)


rjbs:
1.  let's rely on v5.20+ so you can just use glob
2. please put this in DZ::Util as homedir() so we can call it wherever without reimplementing

karenetheridge:
\#2 done; the implementation will become nice and tiny once 5.20 is required.

rjbs:
Thanks, applied in my master!

PR #606, opened 2017-12-23
Fix case of 'CPANFile' in example POD

dboehmer:
Adding 'CPANfile' to dist.ini doesn't work.


rjbs:
I applied an equivalent MR, thanks!

PR #616, opened 2018-02-25
do not add Archive::Tar::Wrapper as a required prereq

karenetheridge:
This mistakenly occurred as a result of commit ff42620c.


rjbs:
Thanks, applied!

Issue #612, opened 2018-02-11
6.011: t/plugins/filefinders.t fails (with older File::Copy::Recursive?)


eserte:
On some of my smokers the test suite fails:
```
#   Failed test 'GatherDir gathers all files in the source dir'
#   at t/plugins/filefinders.t line 52.
# Comparing $data as a Bag
# Missing: 'bin/another_perl_script', 'bin/test.bash', 'bin/test.pl', 'inc/Foo.pm', 'inc/Foo/Bar.pm', 'share/my_data.dat'

#   Failed test 'IncModules finds all modules'
#   at t/plugins/filefinders.t line 86.
# Comparing $data as a Bag
# Missing: 'inc/Foo.pm', 'inc/Foo/Bar.pm'

#   Failed test 'ExecFiles finds all files'
#   at t/plugins/filefinders.t line 113.
# Comparing $data as a Bag
# Missing: 'bin/another_perl_script', 'bin/test.bash', 'bin/test.pl'

#   Failed test 'PerlExecFiles finds exec files that are perl'
#   at t/plugins/filefinders.t line 124.
# Comparing $data as a Bag
# Missing: 'bin/another_perl_script', 'bin/test.pl'

#   Failed test 'ShareFiles finds all files'
#   at t/plugins/filefinders.t line 134.
# Comparing $data as a Bag
# Missing: 'share/my_data.dat'
# Looks like you failed 5 tests of 12.
t/plugins/filefinders.t ......... 
Dubious, test returned 5 (wstat 1280, 0x500)
Failed 5/12 subtests 
```

Statistical analysis suggests that this happens with File::Copy::Recursive older than 0.40:
```
****************************************************************
Regression 'mod:File::Copy::Recursive'
****************************************************************
Name           	       Theta	      StdErr	 T-stat
[0='const']    	      1.0000	      0.0000	49808764893939896.00
[1='eq_0.39']  	     -1.0000	      0.0000	-19388035439960864.00
[2='eq_0.40']  	      0.0000	      0.0000	   1.60

R^2= 1.000, N= 36, K= 3
****************************************************************
```
Please also note that FCR 0.40 currently fails on most BSD and Windows systems:
http://fast-matrix.cpantesters.org/?dist=File-Copy-Recursive%200.40

karenetheridge:
This is also #609.

eserte:
It does not seem that the prereq version was increased in 6.012:
```
Prerequisite modules loaded:
...
[...]

Issue #609, opened 2017-12-30
t/plugins/filefinders.t fails under brewed 5.26.1 on Ubuntu 17.10 64bit


abraxxa:
From the `cpanplus i Dist::Zilla` output:

```
t/plugins/encoding.t ............ ok
t/plugins/extratests.t .......... ok
t/plugins/fakerelease.t ......... ok
t/plugins/ffbyname.t ............ ok

#   Failed test 'GatherDir gathers all files in the source dir'
#   at t/plugins/filefinders.t line 52.
# Comparing $data as a Bag
# Missing: 'bin/another_perl_script', 'bin/test.bash', 'bin/test.pl', 'inc/Foo.pm', 'inc/Foo/Bar.pm', 'share/my_data.dat'

#   Failed test 'IncModules finds all modules'
#   at t/plugins/filefinders.t line 86.
# Comparing $data as a Bag
# Missing: 'inc/Foo.pm', 'inc/Foo/Bar.pm'

#   Failed test 'ExecFiles finds all files'
#   at t/plugins/filefinders.t line 113.
# Comparing $data as a Bag
# Missing: 'bin/another_perl_script', 'bin/test.bash', 'bin/test.pl'

#   Failed test 'PerlExecFiles finds exec files that are perl'
#   at t/plugins/filefinders.t line 124.
# Comparing $data as a Bag
# Missing: 'bin/another_perl_script', 'bin/test.pl'

#   Failed test 'ShareFiles finds all files'
#   at t/plugins/filefinders.t line 134.
# Comparing $data as a Bag
# Missing: 'share/my_data.dat'
# Looks like you failed 5 tests of 12.
t/plugins/filefinders.t ......... 
Dubious, test returned 5 (wstat 1280, 0x500)
Failed 5/12 subtests 
t/plugins/filter-b.t ............ ok
t/plugins/gatherdir.t ........... ok
t/plugins/gatherfile.t .......... ok
```

Output of `perl -V`:
```
[alex@alex-pc:~/perl5/git/Test-Portability-Files$ (master)]$ perl -V
Summary of my perl5 (revision 5 version 26 subversion 1) configuration:
   
  Platform:
    osname=linux
    osvers=4.13.0-21-generic
    archname=x86_64-linux
    uname='linux alex-pc 4.13.0-21-generic #24-ubuntu smp mon dec 18 17:29:16 utc 2017 x86_64 x86_64 x86_64 gnulinux '
[...]

PR #617, opened 2018-02-26
Add File::Copy::Recursive version to avoid tests fail

Anufant:
Test for GatherDir fail if you have File::Copy::Recursive@0.39
installed


Anufant:
This PR is for issues #609  and #612 
And unfortunately, I have already spent 3 hours on investigate this issue :)

karenetheridge:
Unfortunately, this patch is no good, because FRC 0.39 and 0.40 fail utterly on Darwin (and some other architectures too).

Related ticket (no headway seems to be forthcoming on it -- even more unfortunate): https://rt.cpan.org/Ticket/Display.html?id=124166

karenetheridge:
side note -- if we can, we should get off FCR - it is getting a bit too high river for its quality, and many of its features are not needed here. Maybe plain old File::Copy (or the wrapper in Path::Tiny) might be suitable here, or one of the recent replacements floated due to the BBC issue.

rjbs:
Updated to v0.41 for now, anyway.

PR #619, opened 2018-04-21
the release-testing branch no longer exists

karenetheridge:
ಠ_ಠ


karenetheridge:
\#oslo

this fails only because of a missing Changes entry.

rjbs:
Done.

PR #620, opened 2018-04-27
fixes for prototype v7

karenetheridge:
1. symlink path was not absolute  
This is necessary now that we no longer chdir into the build directory.

Found while testing some plugin fixes with the tarballed Dist::Zilla 7.000 prototype -- occasionally when running tests in parallel I saw the symlink removal fail immediately after checking for its existence, suggesting a race condition with other tests that were removing the file in the local repository rather than the file in the tzil tempdir. Closer examination confirmed that the wrong directory was being looked at.

2. archive filename became absolute

This string is passed around as the $archive argument to before_release, release and after_release methods in BeforeRelease, Releaser and AfterRelease plugins.  Instead of now passing an absolute filename, which would force changes to plugins, we now document that this is a relative path and it must be used together with the distribution root if the file is to be read in.


karenetheridge:
Looks like some more work is needed to handle the experimental warnings in 5.20 and 5.22. The 5.18 failures look even more interesting!

rjbs:
I think I need to just abandon the v7 branch and decide what, if anything, I can accomplish from that list.  I apologize for the inconvenience. :(

Issue #586, opened 2017-04-16
fails 3 tests with 'Cannot determine local time zone'


kentfredric:
I have a chroot setup where everything else seems fine, but this is fatal:
```
t/plugins/encoding.t ............ ok                                                            
===(     211;12  4/?  0/? )=============================================Cannot determine local time zone
# Tests were run but no plan was declared and done_testing() was not seen.                                           
# Looks like your test exited with 2 just after 4.                                                                       
t/plugins/distmeta.t ............ Dubious, test returned 2 (wstat 512, 0x200)                              
All 4 subtests passed                                                                                                                                     
t/plugins/extratests.t .......... ok                                    
t/plugins/ffbyname.t ............ ok   
t/plugins/makemaker.t ........... ok                                
===(     281;19  11/?  1/? )============================================    # No tests run!
                                                                                                                                
#   Failed test 'No tests run for subtest "non-ASCII author"'       
#   at t/plugins/license.t line 50.                                 
Cannot determine local time zone                                    
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 255 just after 2.                
t/plugins/license.t ............. Dubious, test returned 255 (wstat 65280, 0xff00)              
Failed 1/2 subtests                                                                            
t/plugins/installdirs.t ......... ok                                                                                 
t/plugins/metanoindex.t ......... ok                                                                                     
t/plugins/manifest.t ............ ok                                                                       
t/plugins/metaresources.t ....... ok                                                    
t/plugins/misctests.t ........... ok                                    
t/plugins/modulebuild.t ......... ok                                
t/plugins/munger-that-prunes.t .. ok                                
t/plugins/pkgdist.t ............. ok                                
t/plugins/pkgversion.t .......... ok                                
t/plugins/nextrelease.t ......... ok                                
t/plugins/podsyntaxtests.t ...... ok                                
t/plugins/podversion.t .......... ok                                
t/plugins/prereqs.t ............. ok
t/plugins/release_status.t ...... ok
t/plugins/prunes.t .............. ok
===(     432;29  3/?  0/? )=============================================Cannot determine local time zone
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 2 just after 3.
t/plugins/readme.t .............. Dubious, test returned 2 (wstat 512, 0x200)
All 3 subtests passed
[...]

PR #622, opened 2018-05-19
Explicitly set $ENV{TZ} in three test files.

jkeenan:
As per usage in t/plugins/nextrelease.t.

When testing Build::Zilla in a FreeBSD-11 virtual machine, failures in
three test files were encountered.  In each case, the failure occurred
when '$tzil->build;' was called in a block where the file being
processed contained non-ASCII text.  In each case, an exception was
thrown whose text begins: "Cannot determine local time zone".  This
message is believed to come from DateTime::TimeZone::Local.
(https://grep.metacpan.org/search?qci=&q=Cannot%20determine%20local%20time%20zone&qft=&qd=DateTime-TimeZone&f=lib%2FDateTime%2FTimeZone%2FLocal.pm)

For: https://github.com/rjbs/Dist-Zilla/issues/586

### DIFF
--- a/t/plugins/distmeta.t
+++ b/t/plugins/distmeta.t
@@ -9,6 +9,8 @@ use Test::DZil;
 use JSON::MaybeXS;
 use YAML::Tiny;
 
+local $ENV{TZ} = 'America/New_York';
+
 {
   # 2.0
   my $tzil = Builder->from_config(
--- a/t/plugins/license.t
+++ b/t/plugins/license.t
@@ -6,6 +6,8 @@ use utf8;
 use autodie;
 use Test::DZil;
 
+local $ENV{TZ} = 'America/New_York';
+
 subtest "ASCII-only author" => sub {
   my $tzil = Builder->from_config(
     { dist_root => 'corpus/dist/DZT' },
--- a/t/plugins/readme.t
+++ b/t/plugins/readme.t
@@ -6,6 +6,8 @@ use utf8;
 use autodie;
 use Test::DZil;
 
+local $ENV{TZ} = 'America/New_York';
+
 my $tzil = Builder->from_config(
   { dist_root => 'corpus/dist/DZT' },
   {


ap:
I get the same DateTime::TimeZone::Local exception with `TZ` set to `America/New_York`. Setting it to `Etc/GMT` or `Etc/UTC` works for me.

rjbs:
I am happy to apply _something_ but this feels a bit "kick it until it works."

Perhaps a better change will be to make all DZ tests set TZ unless DateTime is able to detect the time correctly.  Thoughts?

jkeenan:
> I am happy to apply _something_ but this feels a bit "kick it until it works."
> 
> Perhaps a better change will be to make all DZ tests set TZ unless DateTime is able to detect the time correctly. Thoughts?

I have never looked into the code in the Dist::Zilla test suite, but I suspect the number of files in that suite is large -- which implies changes in more than just 3 files.

Also, how would you determine that DateTime is unable to detect the time correctly?

Thank you very much.
Jim Keenan

karenetheridge:
> I have never looked into the code in the Dist::Zilla test suite, but I suspect the number of files in that suite is large -- which implies changes in more than just 3 files.

Not if it was done in Dist::Zilla::Tester.

ap:
> Thoughts?

Sounds reasonable.

rjbs:
@ap Do you still see this problem in the wild?  Jim says he doesn't on his VM.  If so, does #649 address it?

ap:
I can no longer reproduce this problem just by attempting to install Dist::Zilla,. Not sure why. But I verified the patch anyway (see the PR for details) because recent DateTime::TimeZone versions say [you should be doing this regardless](https://metacpan.org/pod/DateTime::TimeZone#The-%22local%22-time-zone).

rjbs:
Should've been fixed in v6.015

PR #624, opened 2018-06-15
Correct documented default format for NextRelease

Grinnz:


karenetheridge:
note to the maintainer: this patch is suitable for applying to v6, not just v7.

Grinnz:
Rebased to current master

rjbs:
Applied.

PR #625, opened 2018-06-18
Some MakeMaker fixes for underscores in ExtUtils::MakeMaker $VERSION

karenetheridge:
These reflect changes released in Dist::Zilla::Plugin::MakeMaker::Awesome 0.43, 0.44 and 0.45.

- ensure that version checks against eumm_version work when ExtUtils::MakeMaker itself had an underscore in its $VERSION (which occurs in at least one stable perl)
- ensure that static version checks also do the right thing against such EUMM versions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ perl:
   - "5.16"
   - "5.14"
 before_install:
-  - git clone git://github.com/rjbs/travis-perl-helpers -b release-testing ~/travis-perl-helpers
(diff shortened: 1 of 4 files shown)


karenetheridge:
I based this branch against maint-v6. I still think that the v7 work should be moved to a side branch, and v6 moved back to master, until 7.000-TRIAL is released (it isn't ready to be released yet and the timeframe looks quite hazy given scheduling in meatspace).

Grinnz:
It is unnecessary to arbitrarily remove underscores here. Version formats should be preserved.

PR #626, opened 2018-06-21
Move TRIAL comment to end of line when use_begin enabled

nrdvana:
When use_begin is enabled, and the version number is a trial,
the result is 'BEGIN { ... # TRIAL };' which is a compile
error.  This rearranges the code slightly to move the comment
to the end of the line.


karenetheridge:
#v6

nrdvana:
I rebased, but PR still looks like a mess.  I'll create a new clean one.

karenetheridge:
> I rebased, but PR still looks like a mess.

Looks like you brought some extra v7 commits with you when you pushed. You'd need to remove those from the rebase as well.

PR #627, opened 2018-07-02
also add . to @INC for "dzil authordeps --missing"

karenetheridge:
Dist::Zilla 6.012 is compatible with no-.-in-@INC, with the exception of `dzil authordeps --missing`, which still fails to find local plugins.  Here we add it in.

As this fixes an existing bug with modern perls, rather than adding new features, it is suitable for merge in the v6 series.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,8 @@
 Revision history for {{$dist->name}}
 
 {{$NEXT}}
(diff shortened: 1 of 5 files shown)


karenetheridge:
rebased to new master.

rjbs:
Thanks, applied!

PR #629, opened 2018-08-12
add header to cpanfile saying it was generated

preaction:
I keep getting PRs for my cpanfile that I keep in the root of my
distribution. To try to head this off, this commit adds a header to the
generated cpanfile telling users that editing it won't work. This
comment is configurable from the dist.ini file


karenetheridge:
+1. IMHO all auto-generated files should have a comment header explaining where the code lives that generated it (ideally plugin name plus version).

preaction:
I agree with adding a version, though I couldn't find a `$VERSION` variable to use that would make the tests pass (since, I suspect, they're added automatically during the build process). If there's a way I can make that work, let me know and I can update this PR.

karenetheridge:
I usually do `$VERSION // '<internal>'` for distributions that don't have a version declaration right in the repo.

preaction:
Done, the header now includes the version of the plugin.

preaction:
I've reset my branch to begin from the current master.

rjbs:
Thanks, applied!

PR #634, opened 2018-11-27
Add SPDX expression to the metadata

Leont:
This adds an [SPDX](https://en.wikipedia.org/wiki/Software_Package_Data_Exchange) license expression to the metadata. This allows better signaling of licenses to downstream users and vendors.


karenetheridge:
IMO this is a good change and should go into v6.

Issue #632, opened 2018-10-05
With dzil run command -x  flags are eaten by dzil


ambs:
`dzil run prove -v -b t/some-test` vs `dzil run 'prove -v -b t/some-test'`
In the first, -v and -b are interpreted as flags for dzil, and it shouldn't (or at least, it would be nice to).

karenetheridge:
This is pretty standard for shell things -- try `dzil run -- prove -v -b t/some-test`.

ambs:
True. I just though it would be good to report and suggest. Happy if you wish to just close the issue.

jonasbn:
I can confirm the observation by @ambs

jonasbn:
Commented lifted from PR #635 

> @Leont I understand your standpoint, the suggested fix is not pretty. Could you please elaborate on what you mean by permutation and how you would approach eliminating the use of --, since it is a common usage pattern.

Leont:
'permute' is the getopt option that allows it to not stop at the first non-option (in this case 'prove'), but skip over it and only consume the options (and their arguments).

What I don't understand is how the getopt doesn't consume the `--`. `App::Cmd` does some pretty complex things to parse arguments.

PR #635, opened 2018-12-15
First shot at fix for issue #632

jonasbn:
Hard to fix something, which does not seem to have a proper name and convention.

> Two hyphen-minus characters without following letters (--) may indicate that the remaining arguments should not be treated as options, which is useful for example if a file name itself begins with a hyphen,or if further arguments are meant for an inner command (e.g., sudo).

Ref: https://en.wikipedia.org/wiki/Command-line_interface#Option_conventions_in_Unix-like_systems

### DIFF
--- a/lib/Dist/Zilla/Dist/Builder.pm
+++ b/lib/Dist/Zilla/Dist/Builder.pm
@@ -802,6 +802,13 @@ non-zero, the directory will be left in place.
 sub run_in_build {
   my ($self, $cmd, $arg) = @_;
 
+
+  # eliminate the no-options parameter, since indicating this as a first
+  # parameter does not work
+  if ($cmd->[0] eq '--') {
+    shift @{$cmd};
+  }
+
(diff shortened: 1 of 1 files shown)


jonasbn:
@Leont I understand your standpoint, the suggested fix is not _pretty_. Could you please elaborate on what you mean by permutation and how you would approach eliminating the use of `--`, since it is a common usage pattern.

jonasbn:
Fundamentally the issue seems to be with the call to `system` and feeding the `--` as first argument and first argument should be the actual command. Fixing this in [App::CMD](https://metacpan.org/pod/App::Cmd) could be an interesting and perhaps better approach. So the suggestion would be to add an feature to eliminate the `--` via App::CMD's processing prior to the call to `system` in Dist::Zilla?

jonasbn:
hi @Leont 

I will kill this PR and possibly investigate an alternative way to address the problem

Take care,

jonasbn

PR #636, opened 2018-12-16
v7 only supports perl 5.20+

karenetheridge:
This will get travis builds passing again on branches based off of v7-in-progress.


rjbs:
Thanks, applied.  I'm not sure whether v7 is how I'd proceed on v7 now, but … dunno!

PR #637, opened 2018-12-16
remove all uses of File::HomeDir

karenetheridge:
...as inspired by File::HomeDir::Tiny

Note that all of this can be replaced by (glob('~'))[0] when perl 5.20
is required.

(This is a simplification of a piece of an earlier pull request, and is suitable for merging in the v6 series.)

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,6 +1,7 @@
 Revision history for {{$dist->name}}
 
 {{$NEXT}}
+        - replace uses of File::HomeDir with a simple glob()
 
 6.012     2018-04-21 10:20:21+02:00 Europe/Oslo
         - revert addition of Archive::Tar::Wrapper as a mandatory prereq (in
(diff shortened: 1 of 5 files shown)


Leont:
> Note that all of this can be replaced by (glob('~'))[0] when perl 5.20
is required.

Actually, that only requires 5.14 AFAIK (if you need it to be functional on Windows, on other platforms it has worked since time immemorial)

karenetheridge:
Yes, but Dist::Zilla does support windows :)

Leont:
> Yes, but Dist::Zilla does support windows :)

Yeah, but it hasn't supported <5.14 in ages

karenetheridge:
Windows on 5.14 would still need the special treatment. It's only in v7, where the minimum version rises to 5.20, that we can drop the extra code.

karenetheridge:
This is safe to apply as-is in v6, should you wish it (we did a simplified version in v7 that drops the version check).

rjbs:
c'est merged

Issue #621, opened 2018-05-13
Dist::Zilla should provide a way to globally set parallel job execution


autarch:
Right now it's impossible to get `dzil release` to run tests in parallel short of subclassing the TestRunner plugin (and RunExtraTests if you're using that) and overriding the `default_jobs` method.

It'd be really nice if dzil supported a single global way to set the jobs for all things that can run in parallel. IMO, it'd be even better if it detected the number of cores on the system and just run `$cores * 2` jobs as a default. Or maybe this could be the default when `-j` is passed without a number.

Leont:
`HARNESS_OPTIONS=j8 dzil test`?

karenetheridge:
FWIW I pass the `default_jobs` option through to my MakeMaker and ModuleBuildTiny plugins in my bundle.

Leont:
> HARNESS_OPTIONS=j8 dzil test?

Nope, that doesn't work because dzil overrides it even when no default_jobs is set. That is unfortunate.

Adding a `jobs` argument to `release` shouldn't be too hard though.

> FWIW I pass the default_jobs option through to my MakeMaker and ModuleBuildTiny plugins in my bundle.

IMO that should be up to the user running it.

karenetheridge:
> IMO that should be up to the user running it.

Agreed.  Ultimately, this is one of the configs that I'd want to put in ~/.dzil/config.ini, if there was better support for this. It doesn't belong in a bundle or in a plugin override.

Leont:
I created #638 to fix this :-)

rjbs:
This was fixed as it says above (6.014)

PR #638, opened 2018-12-22
Add a --jobs argument to dzil release #621

Leont:
This adds a --jobs argument to dzil release. Along the way it also fixes the mindless overriding of HARNESS_OPTIONS.


rjbs:
Thanks, applied and will release soon.

Issue #646, opened 2019-12-01
Abstract extraction cannot handle escapes


davel:
Dist::Zilla doesn't decode escaped characters when it extracts the abstract from POD. The escape sequences end up in META.*.

For example,
```
=head1 NAME

Smiling::Cat::With::Heart::Shaped::Eyes - Looks like E<128571>

=cut
```
yields `abstract: ' Looks like E<128571>'` rather than `abstract: 'Looks like 😻'`.

karenetheridge:
By any chance have you noticed a published distribution on cpan where this problem can be observed?

davel:
> By any chance have you noticed a published distribution on cpan where this problem can be observed?

Not AFAIK. In this one I corrected the abstract in dist.ini; https://metacpan.org/release/DLAMBLEY/App-RetroPAN-0.01

PR #647, opened 2019-12-01
Decode escaped POD chars in abstract

davel:
This fixes https://github.com/rjbs/Dist-Zilla/issues/646.

I also considered sending a patch for `Pod::Eventual` to resolve this. However that makes no attempt to even parse escape characters, and so risks causing weird bugs in users which already do their own decoding.


karenetheridge:
I wonder if it should do more parsing than that? e.g. what if the abstract line was `This is my B<awesome> module` -- we would want to strip that bold tag out as well.  There should be a function in one of the pod libraries that gives us the literal text contained in a pod string.

davel:
I had a search through CPAN but could not find a module for extracting abstracts. I could pull the code out into a new module if you'd be happy to add a dependency?

MetaCPAN does something similar, but gets escaped characters wrong in a different manner (it also looks like it doesn't extract abstracts from code comments, unlike Dist::Zilla), https://github.com/metacpan/metacpan-api/blob/master/lib/MetaCPAN/Document/File.pm#L110

There is perhaps less of a use case for handling mark-up other than `E<...>` inside abstracts, but I agree that it would be desirable to handle it consistently. I suspect there's few if any modules with mark-up in the abstract, if only because the tools cannot handle it.

karenetheridge:
> but could not find a module for extracting abstracts

I would not expect one for abstracts specifically. A pod->text converter (i.e. a pod function that strips all markup entirely either from a document section or a string would be just fine here -- e.g. Pod::Text.

> it also looks like it doesn't extract abstracts from code comments, unlike Dist::Zilla

Well no, it wouldn't, because metacpan is starting with a distribution that has already been uploaded to PAUSE, which is the very thing that Dist::Zilla is generating.  But it *should* handle pod markup properly (and ideally preserve it for rendering in html, rather than stripping out entirely) -- so it ought to run the content through Pod::Text or Pod::HTML.

Grinnz:
I use Pod::Simple::TextContent for instances like this.

davel:
Thank you for the suggestions. I don't think we can pass the text through `Pod::Text` directly, because (to keep existing tested behaviour) the `C<...>` escape needs to be preserved. An example in the wild is https://metacpan.org/release/FLORA/Term-ReadLine-1.14

I have therefore switched to use Pod::Simple (which Pod::Text uses), for the full parse (rather than using `Pod::Eventual` for the main structure and `Pod::Simple` for the escapes). This has the advantage of respecting character encoding of the source the same way that other POD tools do.

lib/Dist/Zilla/Util.pm
```diff
@@ -9,45 +9,73 @@ use Encode ();
...
+  @Dist::Zilla::Util::PEA::ISA = ('Pod::Simple');
+
   sub _new  {
-    # Load Pod::Eventual only when used (and not yet loaded)
-    unless (exists $INC{'Pod/Eventual.pm'}) {
-      require Pod::Eventual;
-      Pod::Eventual->VERSION(0.091480); # better nonpod/blank events
+    my ($class, @args) = @_;
+    # Load Pod::Simple only when used (and not yet loaded)
+    unless (exists $INC{'Pod/Simple.pm'}) {
```
rjbs: require is a no-op when a library is already loaded, so this branch is unneeded

rjbs:
Thanks!  I have applied this and will include it in a trial release Very Soon™.

In the future, please include a Changes entry, which will make my life just a wee bit easier.

davel:
Closing, had not noticed the Pod::Simple changes were already merged.

be88369d65b3603c4201f212055a6a5bda8552ed fixes the comment, but not really part of this pull request.

PR #649, opened 2020-01-26
Tester: locally set $ENV{TZ} if it seems needed

rjbs:
I think this fixes #622

### DIFF
--- a/lib/Dist/Zilla/Tester.pm
+++ b/lib/Dist/Zilla/Tester.pm
@@ -205,6 +205,13 @@ sub minter { 'Dist::Zilla::Tester::_Minter' }
   around build_in => sub {
     my ($orig, $self, $target) = @_;
 
+    # Sometimes, we can't get a local time zone.  When that happens, we're just
+    # going to pretend to be in UTC.  We don't do this during actual runtime
+    # because the user can fix their own environment, but we'll let them do
+    # that after they get the software installed. -- rjbs, 2020-01-26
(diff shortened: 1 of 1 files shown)


jkeenan:
@rjbs, I would say: go ahead and merge this and close my original p.r..

Unfortunately, I cannot replicate the bug I reported in https://github.com/rjbs/Dist-Zilla/pull/622.  My original report states that the problem was observed in a FreeBSD-11 VM.  The last time I used either of those VMs was last June.  I just restarted both of them.  The current Dist::Zilla installs fine in both.  So I have no evidence that your patch does not correct the original problem.

I suspect that what happened was that, subsequent to filing the original report, I did some sysadmin-ish work in the VM that enabled DateTime::TimeZone::Local to correctly determine the TZ.

Thank you very much.
JIm Keenan

ap:
Looks good to me.

To be able to verify the patch regardless of environment, I patched DateTime::TimeZone::Local::Unix to reduce `sub Methods` to `return 'FromEnv'`. This prevents fallback to the non-`$ENV{TZ}` detection methods. This renders DateTime::TimeZone::Local unable to detect the local timezone unless  `TZ` is set.

I ran the Dist::Zilla tests in this environment, and as expected, the exact same tests mentioned in #622 failed with the exact same error.

I then applied this patch, re-ran the tests, and they passed.

So this patch does what it’s supposed to.

rjbs:
Thanks, applied!

PR #650, opened 2020-01-30
fix missing newline before pod when TRIAL and versionfix

Sadrak:
I had a strange error in a special case. The pod would be not starting in a fresh line!

PkgVersion produce something like this:
`$DZT::PodWithPackageTrial::VERSION = '0.004002';=pod`

It also depends on how many newlines are at the beginning before the package definition. When the package is the first line it breaks, when the package is after a empty line everything works.

Attached a test and the fix for this special case.


Sadrak:
Perhaps an easier solution is to only move the line which adds the newsline after the clean_version statement.

rjbs:
Thanks, applied!

rjbs:
…and released.

PR #651, opened 2020-02-04
require is a no-op if package is already loaded

davel:
Fixes https://github.com/rjbs/Dist-Zilla/pull/647#pullrequestreview-348331817 as suggested.


rjbs:
Applied, despite first bad attempt to fix conflict (on my part).

PR #652, opened 2020-05-02
Added some basic documentation of the jobs option

jonasbn:
I was updating the Dash cheatsheet and saw that the documentation had not been updated with the `--jobs` option.

So here is a suggestion

### DIFF
--- a/lib/Dist/Zilla/App/Command/release.pm
+++ b/lib/Dist/Zilla/App/Command/release.pm
@@ -11,10 +11,31 @@ use Dist::Zilla::App -command;
 
   dzil release --trial
 
+  # long form, jobs takes an integer
+  dzil release --jobs 9
+
+  # short form, same as above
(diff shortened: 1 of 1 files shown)


lib/Dist/Zilla/App/Command/release.pm
```diff
@@ -11,10 +11,31 @@ use Dist::Zilla::App -command;
 
   dzil release --trial
 
+  # long form, jobs takes an integer
+  dzil release --jobs 9
+
+  # short form, same as above
+  dzil release --jobs 9
```
oalders, line 18: Should the short form be `dzil release -j 9`?

jonasbn: Oops, yes

jonasbn: Fixed accordingly

PR #653, opened 2020-05-22
restore Changes entry for 6.013

karenetheridge:
looks like a bad rebase of aae06736 caused this section to be accidentally lost.


rjbs:
Thanks, it's back.!

Issue #654, opened 2020-05-23
`dist.pl` support seems entirely broken.


kentfredric:
```
/dist-zilla/corpus/dist/DZ2 $ dzil build
Use of uninitialized value $str in index at /home/kent/perl5/perlbrew/perls/5.27.9-nossp-oaath-nopmc-nodot/lib/site_perl/5.27.9/String/RewritePrefix.pm line 79.
Use of uninitialized value $str in index at /home/kent/perl5/perlbrew/perls/5.27.9-nossp-oaath-nopmc-nodot/lib/site_perl/5.27.9/String/RewritePrefix.pm line 79.
Use of uninitialized value $str in index at /home/kent/perl5/perlbrew/perls/5.27.9-nossp-oaath-nopmc-nodot/lib/site_perl/5.27.9/String/RewritePrefix.pm line 79.
Use of uninitialized value $str in index at /home/kent/perl5/perlbrew/perls/5.27.9-nossp-oaath-nopmc-nodot/lib/site_perl/5.27.9/String/RewritePrefix.pm line 79.
Use of uninitialized value $str in substr at /home/kent/perl5/perlbrew/perls/5.27.9-nossp-oaath-nopmc-nodot/lib/site_perl/5.27.9/String/RewritePrefix.pm line 85.
Attribute (name) does not pass the type constraint because: Validation failed for 'Str' with value undef at constructor Dist::Zilla::MVP::Section::new (defined at /home/kent/perl5/perlbrew/perls/5.27.9-nossp-oaath-nopmc-nodot/lib/site_perl/5.27.9/Dist/Zilla/MVP/Section.pm line 44) line 61
	Dist::Zilla::MVP::Section::new('Dist::Zilla::MVP::Section', 'HASH(0x55e4ba577480)') called at /home/kent/perl5/perlbrew/perls/5.27.9-nossp-oaath-nopmc-nodot/lib/site_perl/5.27.9/Config/MVP/Assembler.pm line 144
	Config::MVP::Assembler::begin_section('Dist::Zilla::MVP::Assembler::Zilla=HASH(0x55e4b7bc30d0)', undef, undef) called at /home/kent/perl5/perlbrew/perls/5.27.9-nossp-oaath-nopmc-nodot/lib/site_perl/5.27.9/Config/MVP/Assembler.pm line 187
[...]

PR #655, opened 2020-05-23
Fix loading of `dist.pl` re GH#654

kentfredric:
This provides the most minimal changes I think might work.

( Though I suspect more changes are desired/necessary to bring full feature-comparability with the likes of dist.ini and bundles )

Importantly:

- The sub called $self->assembler instead of using the passed $asm,
  indicating that this code probably never worked.

- When a plugin definition is not specified as a ( key => ref ) pair,
  and instead there is only 'key', if the next element doesn't exist,
  the existing code shoves an `undef` back into @plugins when it should
  do nothing, resulting in a complete failure trying to create a section
  with an undef name, entirely breaking things like:

    [ '@Classic' ]

  This code no longer injects a new "undef" that was previously not in
  the input data, into the plugin data.

- A very very rudimentary 'it builds' test is added for DZ2/dist.pl test
  case.

Bug: https://github.com/rjbs/Dist-Zilla/issues/654

### DIFF
--- a/t/distpl.t
+++ b/t/distpl.t
@@ -1,0 +1,14 @@
+use strict;
+use warnings;
+use Test::More 0.88;
+
+use Test::DZil;
+
+my $tzil = Builder->from_config(
+  { dist_root => 'corpus/dist/DZ2' },
+);
+
+$tzil->chrome->logger->set_debug(1);
+$tzil->build;
+pass();
+done_testing;
--- a/lib/Dist/Zilla/MVP/Reader/Perl.pm
+++ b/lib/Dist/Zilla/MVP/Reader/Perl.pm
@@ -30,7 +30,7 @@ sub read_into_assembler {
 
   while (my ($ident, $arg) = splice @$plugins, 0, 2) {
     unless (ref $arg) {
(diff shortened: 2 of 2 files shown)


kentfredric:
Pleased to report with this change, the following horror show works:

https://github.com/kentfredric/Dist-Zilla-Plugin-MetaConfig-External/blob/master/dist.pl

Yep, keeping all the logic in one file is possible :grinning: 

Though I probably won't do exactly this, ... its just nice to know I can.

kentfredric:
Caveats: authordeps can't work. *wah wah*

t/distpl.t
```diff
@@ -0,0 +1,16 @@
+use strict;
+use warnings;
+use Test::More 0.88;
+
+use ExtUtils::Manifest 'maniread';
```
karenetheridge, line 5: is this needed?

kentfredric: mm, no, I just cribbed the code semi-blindly from another test and then cut it down till it worked, clearly didn't cut it enough. :)

rjbs:
Thanks, applied.

PR #661, opened 2020-08-11
Add prerequisite App::cpanminus

molecules:
After installing Dist::Zilla via `cpan`, I tried running `dzil install` and it failed with the following error:

>    Can't exec "cpanm": No such file or directory 

After installing App::cpanminus, there were no problems, hence the suggested addition of App::cpanminus as a dependency.

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -46,6 +46,8 @@ Mixin::Linewise::Readers = 0.100    ; default encodings to UTF-8
 
 DateTime = 0.44 ; CLDR fixes, used by AutoVersion and NextRelease
 
+App::cpanminus = 0; Essentail for dzil install  
+
 [OSPrereqs / MSWin32]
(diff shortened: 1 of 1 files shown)


karenetheridge:
IMO this should be a RuntimeRecommends, not a required prereq.

molecules:
@karenetheridge Is there an option to use `cpan` instead of `cpanm` for `dzil install`?

karenetheridge:
```
$ dzil help install
dzil install [long options...]

	--install-command STR    command to run to install (e.g. "cpan .")
	--keep --keep-build-dir  keep the build directory even after a success

```

molecules:
@karenetheridge Nice! Okay, I made the suggested change. Thanks!

rjbs:
Thanks, applied!

PR #666, opened 2021-04-06
fix PkgVersion using use_our and underscore versions

haarg:
Previously, a version using an underscore with the use_our option set
would generate code like:

```
{ our $VERSION = '0.001_001'; } #TRIAL

$VERSION = '0.001001';
```

This won't work under strict. Fix this by factoring out the version
assignment generation code, so it will be kept consistent between the
two assignments. Now the generated code will look like:

```
{ our $VERSION = '0.001_001'; } #TRIAL
{ our $VERSION = '0.001001'; }
```

Also cleans up adding an extra newline between the lines in many cases,
and eliminates an extra set of braces when combining use_our with
use_begin.

### DIFF
--- a/lib/Dist/Zilla/Plugin/PkgVersion.pm
+++ b/lib/Dist/Zilla/Plugin/PkgVersion.pm
@@ -169,6 +169,22 @@ has use_begin => (
   default => 0,
 );
 
+sub _version_assignment {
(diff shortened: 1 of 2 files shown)


rjbs:
Thanks, looks like a clear fix.

Issue #662, opened 2020-09-10
Option to not use Archive::Tar::Wrapper


plicease:
`Archive::Tar::Wrapper` machines with (not particularly) recent version of GNU tar produce tarballs that are not readable by the default tar implementation on HP-UX.

https://github.com/PerlAlien/Alien-Build/issues/228

`Archive::Tar` doesn't seem to have this limitation.  It doesn't seem possible atm to disable `Archive::Tar::Wrapper` from`dist.ini` or (preferably) from a plugin bundle without mucking with the `Dist::Zilla` internals.  This is a request to make this possible without resorting to trickery like `Devel::Hide` or somesuch.

plicease:
Another option would obviously be to update `Archive::Tar::Wrapper` to either produce a compliant tarball or to add an interface to do the same.

https://github.com/glasswalk3r/archive-tar-wrapper-perl/issues/18

plicease:
I think this can be closed now.  Although there is not a simple "turn off use of Archive::Tar::Wrapper" option, the ArchiveBuilder role can be used to write plugins that effectively avoid its usage.

PR #668, opened 2021-06-14
Add roles for plugins to create their own archives for the dist

plicease:
It seems surprising that there is no way to customize the way tarballs are created with `Dist::Zilla` given how customizable it otherwise is in other areas.  I ran into #662 when somebody complained that they weren't able to install one of my modules on HP-UX with a very old version of tar.  Since this is understandably not a priority for others I wrote this patch to add some hooks to allow customize the tarball creation.

Currently I am disabling `Archive::Tar::Wrapper` by patching it https://metacpan.org/dist/Dist-Zilla-PluginBundle-Author-Plicease/source/lib/Dist/Zilla/PluginBundle/Author/Plicease.pm#L66-69 but that would be a bad way to do it, if there were a public interface for doing this.

I'm open to changing the naming of these roles and methods, and I haven't yet written any documentation or tests, but plan to if this approach is broadly acceptable.  I have tested the usefulness of this interface with a proof-of-concept plugin:

https://github.com/uperl/Dist-Zilla-Plugin-Libarchive/blob/e7782e07599d6ebef4991e43da27945ef410b765/lib/Dist/Zilla/Plugin/Libarchive.pm

I appreciate your consideration and your feedback.

Todo:
  - [x] proof of concept works with plugin
  - [x] Document roles
  - [x] Write tests


lib/Dist/Zilla/Dist/Builder.pm
```diff
@@ -462,11 +462,28 @@ might not exist.
 
 sub archive_filename {
   my ($self) = @_;
-  return join(q{},
+
+  my $name = join(q{},
     $self->dist_basename,
     ( $self->is_trial && $self->version !~ /_/ ? '-TRIAL' : '' ),
-    '.tar.gz'
   );
```
rjbs, line 469: Just an FYI while I'm reading this as a proof of concept:  when lines don't match style, I am likely to reformat or rework them to do so.  Here, for example, the use of `(...)` on join is not standard.  This is more a "hey expect this to happen" than a "I am grumpy when things are not exactly how I like them on my plate!

lib/Dist/Zilla/Dist/Builder.pm
```diff
@@ -462,11 +462,28 @@ might not exist.
...
+
+  for my $namer ($self->plugins_with(-ArchiveNamer)->@*) {
+    my $this_filename = $namer->archive_filename($name);
+
+    next unless defined $this_filename;
+
+    $self->log_fatal('attempted to set archive_filename twice')
+      if defined $provided_filename;
+
+    $provided_filename = $this_filename;
```
rjbs, line 481: When you do this, you should also capture the plugin name.  Then, when you hit the log_fatal condition, you can say which two plugins are responsible.

lib/Dist/Zilla/Dist/Builder.pm
```diff
@@ -486,7 +503,12 @@ sub build_archive {
   my $basename = $self->dist_basename;
   my $basedir = path($basename);
 
-  $_->before_archive for @{ $self->plugins_with(-BeforeArchive) };
+  $_->before_archive for $self->plugins_with(-BeforeArchive)->@*;
+
+  for my $builder ($self->plugins_with(-ArchiveBuilder)->@*) {
+    my $file = $builder->build_archive($self->archive_filename, $built_in, $basename, $basedir);
+    return $file if defined $file;
```
rjbs, line 510: On further reading, I think it's "you're not gonna need it" to make an ArchiveNamer plugin.  I think it's just going to be complexity that doesn't get us anything.  Do you have concrete use cases in mind?

Instead, I'd let DZ's Builder code come up with a base name on its own (as it already does), then pass that base name to the ArchiveBuilder which can build the archive and return the full name.

rjbs (changes requested):
I think the idea of an ArchiveBuilder is a good one, and one I might even use!  I think the ArchiveNamer is YAGNI territory, at least at first blush.  Thanks for this start!

lib/Dist/Zilla/Dist/Builder.pm
karenetheridge: If you really want to need to customize the archive name, I'd pass a sprintf-ish style format string to specify the template. (see Dist::Zilla::Plugin::NextRelease and Dist::Zilla::Plugin::Run for other examples of templating short strings.)

plicease: The application initially intended is to set the extension on the archive for alternative formats.  Allowing .zip, for example I think is not desirable tbh for CPAN (although there are some dists that do this), but I also use dzil for stuff not intended for CPAN where this could be useful.  I viewed this approach as less invasive change in my initial implementation, but I agree that computing the extension inside the builder and returning the complete path is better because you could don't have to worry about an ArchiveBuilder+ArchiveNamer competing with another plugin that does ArchiveNamer.  That said, when I re-worked it below to remove the ArchiveNamer, it actually seems pretty non-invasive.

I think the ArchiveNamer might be useful for Perl-dists on non-CPAN type applications, but I don't have a concrete use case for now, and with the iteration below it would probably operate on the archive_basename rather than the full archive_filename anyway.  In the very unlikely situation I run into a need for this I think we can iterate later.

lib/Dist/Zilla/Dist/Builder.pm
```diff
@@ -511,7 +510,7 @@ sub build_archive {
              ? '_build_archive_with_wrapper'
              : '_build_archive';
 
-  my $archive = $self->$method($built_in, $basename, $basedir);
+  my $archive = $self->$method($built_in, $basedir);
```
plicease, line 513: If there is a good reason to pass in the `$basename` I am fine doing that, but as mentioned in the commit I noticed neither Dist::Zilla nor my example plugin (updated here: https://github.com/uperl/Dist-Zilla-Plugin-Libarchive/blob/f5e705a67d765f2299c7b1743de114a22a76b47a/lib/Dist/Zilla/Plugin/Libarchive.pm) were using it, and the `Dist::Zilla` methods are private and it should be fine to change the signature.

lib/Dist/Zilla/Dist/Builder.pm
```diff
@@ -483,17 +496,21 @@ sub build_archive {
...
 
-  my $basename = $self->dist_basename;
-  my $basedir = path($basename);
+  my $basedir = path($self->dist_basename);
 
-  $_->before_archive for @{ $self->plugins_with(-BeforeArchive) };
+  $_->before_archive for $self->plugins_with(-BeforeArchive)->@*;
+
+  for my $builder ($self->plugins_with(-ArchiveBuilder)->@*) {
+    my $file = $builder->build_archive($self->archive_basename, $built_in, $basedir);
```
plicease, line 504: - alternatively we could pass in arguments as a  hash reference
 - could also omit `$self->archive_basename` as an argument; if this is a public method I think it can be called from it from the ArchiveBuilder

plicease: In fact I think all of these arguments could be fetched from the `$zilla` object?

plicease:
Added documentation and test for this.  As mentioned, it might make sense to refine the interface for how or indeed if arguments are passed to this method.  Most roles seem to not pass arguments if they can easily be queried from the `$zilla` instance and that seems to make sense.  On the other hand the three arguments that I picked are going to be necessary I think 99% of the time for real implementations.

rjbs:
Thanks, rebased and merged!

PR #670, opened 2021-06-27
Require PPI 1.222 for package version accessor

### LINKED ISSUE #669: Dist::Zilla depends on any version of PPI but uses PPI::Statement::Package->version which was add in 1.222
When installing Dist::Zilla with Strawberry Perl 5.20.3 in a Github Actions I got the following error message

```
#   Failed test 'No tests run for subtest "use_package"'
##[error]#   at t/plugins/pkgversion.t line 570.
##[error]Can't locate object method "version" via package "PPI::Statement::Package" at C:\HOSTED~1\windows\STRAWB~1\520~1.3\x64\data\.cpanm\work\1624807104.6644\Dist-Zilla-6.020\blib\lib/Dist/Zilla/Plugin/PkgVersion.pm line 249.
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 255 just after 31.
t/plugins/pkgversion.t .......... 
Dubious, test returned 255 (wstat 65280, 0xff00)
Failed 1/31 subtests 
```

The Strawberry Perl in the Github Action comes with PPI 1.220, which lacks the version method in PPI::Statement::Package, which was only added for 1.222 [(Git History)](https://github.com/Perl-Critic/PPI/commits/master/lib/PPI/Statement/Package.pm).

Request is to bump the dependency of Dist::Zilla on PPI to 1.222.

Grinnz:
This module appears to be added to dependencies through autoprereqs currently, but it is lazyloaded so there's no way to auto-set the required version; so just add it manually.

Resolves #669

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -45,6 +45,7 @@ Config::MVP::Reader::INI = 2.101461 ; allow spaces in plugin name
 Data::Section            = 0.200000 ; default encodings to UTF-8
 ExtUtils::Manifest       = 1.54     ; for ManifestSkip that needs maniskip()
 Mixin::Linewise::Readers = 0.100    ; default encodings to UTF-8
+PPI::Document            = 1.222    ; version accessor for package stmt
 Term::ANSIColor          = 5.00     ; 24-bit color support
 
 DateTime = 0.44 ; CLDR fixes, used by AutoVersion and NextRelease


rjbs:
Looks good to me, thanks.

PR #703, opened 2022-01-20
Improve error message for failed 'dzil run'

aquanight:
This improves the error message provided by 'dzil run' to indicate the resulting exit status of a failed command. An example of where this might help is during exploratory testing using 'dzil run perl'.

In particular I ran into this problem when a module I was doing such testing with crashed with SIGSEGV: with no output from perl itself and only a very basic message from dzil, I had to guess (or discover via strace) what had actually happened.

### DIFF
--- a/lib/Dist/Zilla/Dist/Builder.pm
+++ b/lib/Dist/Zilla/Dist/Builder.pm
@@ -20,6 +20,7 @@ use Dist::Zilla::Path; # because more Path::* is better, eh?
 use Try::Tiny;
 use List::Util 1.45 'uniq';
 use Module::Runtime 'require_module';
+use IPC::System::Simple ();
 
 use namespace::autoclean;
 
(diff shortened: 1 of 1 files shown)


karenetheridge:
Can you show a before-and-after of a failed command, since there is no test coverage of this area to show what the changes would look like?

rjbs:
As far as I can tell:

**before**:
```
Can't exec "lss": No such file or directory at /Users/rjbs/.plenv/versions/5.37.2/lib/perl5/site_perl/5.37.2/Dist/Zilla/Dist/Builder.pm line 854.
[DZ] error while running: lss at /Users/rjbs/.plenv/versions/5.37.2/lib/perl5/si
te_perl/5.37.2/Dist/Zilla/Dist/Builder.pm line 854.
[DZ] left failed dist in place at .build/a9IrSHIhsi
```

**after**:
```
[DZ] "lss" failed to start: "No such file or directory" at /Users/rjbs/.plenv/ve
rsions/5.37.2/lib/perl5/site_perl/5.37.2/Dist/Zilla/Dist/Builder.pm line 856.
[DZ] left failed dist in place at .build/tgqXvZr2Xb
```

Is this better?  I don't know, I'm not really seeing the specific way it's better. With no reply to the question already asked, I'm closing this unless there's more to it.

PR #704, opened 2022-02-26
fix "dzil authordeps --missing"

skaji:
https://github.com/rjbs/Dist-Zilla/pull/627 added . to INC for "dzil authordeps --missing".

After that, `dzil authordeps --missing` does not work correctly.

This is because `$root` is an object of `Dist::Zilla::Path` so that `Module::Runtime::require_module` sometimes fails with the message:

```
Can't locate object method "INC" via package "Dist::Zilla::Path"
```

This PR fixes it by stringifying `$root` before adding it to INC.


miyagawa:
+1

rjbs:
Thanks, applied!

PR #706, opened 2022-05-14
suggest cpanm $(dzil) to avoid piping

miyagawa:
Running `dzil ... | cpanm` works but I generally don't recommend it since cpanm loses some functionalities like prompting on failures when STDIN is piped this way.

This PR changes the command outputs to suggest using `cpanm $(dzil ...)` instead.

Note: One subtle difference would be that if the `dzil` commands emit no outputs, the new invocation would give an error message from `cpanm` because no argument would be given to the command. It could be slightly confusing, but I think that's unlikely to happen, since this suggestion is only displayed when there're actually modules missing.

### DIFF
--- a/lib/Dist/Zilla/Dist/Builder.pm
+++ b/lib/Dist/Zilla/Dist/Builder.pm
@@ -314,7 +314,7 @@ Required plugin$bundle $package isn't installed.
 Run 'dzil authordeps' to see a list of all required plugins.
 You can pipe the list to your CPAN client to install or update them:
 
-    dzil authordeps --missing | cpanm
+    cpanm $(dzil authordeps --missing)
 
 END_DIE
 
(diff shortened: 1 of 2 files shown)


haarg:
This won't work properly on Windows.

miyagawa:
Ugh, that's unfortunate. Guess you could detect your shell or OS and suggest accordingly but that'd be tedious.

Curious if we could add a dzil command which would then system/exec `cpanm`, but I guess it might have the same issue if you run via system...

Grinnz:
My `dzil installdeps` command actually works that way, but I can't speak to its portability: https://metacpan.org/pod/Dist::Zilla::App::Command::installdeps

Grinnz:
`dzil install` is also precedent for commands working that way

PR #707, opened 2022-05-14
[MakeMaker] add main_module option

miyagawa:
Add a new option `main_module` to be able to override `NAME` parameter for `WriteMakefile`, to resolve a common case where name of the distribution differs from the main package. e.g.
- https://github.com/libwww-perl/libwww-perl/pull/413
- https://github.com/rjbs/Role-Identifiable/issues/3

According to the LWP PR above, MakeMaker::Awesome seems to allow this, but I think it's common enough that dzil should provide with the default plugin.


rjbs:
Thanks, applied!

PR #708, opened 2022-05-28
convert many subroutines to use signatures

rjbs:


rjbs:
I think this signaturification is not going to happen, or if it does, will happen anew.  Sorry!

PR #710, opened 2022-06-10
Fix build in Windows/msys2

pauloscustodio:
symlink tests must be skipped
trying to remove a directory fails


rjbs:
Thanks, applied!

PR #711, opened 2022-06-19
mark some subroutine arguments as optional

karenetheridge:
This fixes "Too few arguments for subroutine ..." errors from downstream plugins, e.g.:
https://www.cpantesters.org/cpan/report/e5a8a17e-dfa6-11ec-842b-e399c271694f
https://www.cpantesters.org/cpan/report/54d1b008-ec69-11ec-96ae-f2ec22c97367
https://www.cpantesters.org/cpan/report/770212ee-ec69-11ec-be85-cbf022c97367


rjbs:
I think this signaturification is not going to happen, or if it does, will happen anew.  Sorry!

PR #712, opened 2022-06-21
Fix PkgVersion's use-package when multi-dot versioning is used

happy-barney:
then it must be v-string


rjbs:
Thanks!  I amended your commit to keep a comment matching the project style.

Issue #671, opened 2021-07-02
colorized output hard to read on windows


plicease:
On Windows I am seeing the color codes rather than the colors.

```
C:\Users\ollisg\dev\File-XDG>dzil xtest
[←[38;5;199mDZ←[0m] building distribution under .build/9b49CCD1Yt for installation
[←[38;5;199mDZ←[0m] beginning to build File-XDG
[←[38;5;199mDZ←[0m] guessing dist's main_module is lib/File/XDG.pm
[←[38;5;116m@Author::Plicease/Author::Plicease::ReadmeAnyFromPod←[0m] someone tried to munge lib/File/XDG.pm after we read from it. Making modifications again...
[←[38;5;137m@Author::Plicease/Author::Plicease::SpecialPrereqs←[0m] perl version required = 5.006
[←[38;5;199mDZ←[0m] writing File-XDG in .build/9b49CCD1Yt
[←[38;5;106m@Author::Plicease/ReadMePodInRoot←[0m] overriding README.md in root
```

I tried setting DZIL_COLOR=0, but that did not help, I think because of the logic, since when `-t *STDOUT` is true, the environment variable is short-circuited.

PR #713, opened 2022-08-11
allow DZIL_COLOR=0 to turn off colorization regardless of stdout

plicease:
This partly addresses #671 for me.  It makes DZIL_COLOR=0 override the -t test on STDOUT, so that if I set the variable on Windows I can turn off the color output which doesn't work.  This may also benefit non-windows users that want to turn off color output regardless of if this is a run in a terminal.

Arguably colorization should be turned off on windows regardless since ANSI sequences aren't recognized by the cmd.exe window, but I will leave that for a later PR if there is agreement.

### DIFF
--- a/lib/Dist/Zilla/Chrome/Term.pm
+++ b/lib/Dist/Zilla/Chrome/Term.pm
@@ -74,7 +74,7 @@ sub _build_logger {
     quiet_fatal => 'stdout',
   });
 
-  if (-t *STDOUT || $ENV{DZIL_COLOR}) {
+  if (defined $ENV{DZIL_COLOR} ? $ENV{DZIL_COLOR} : -t *STDOUT) {
     my $stdout = $logger->{dispatcher}->output('stdout');
 
     $stdout->add_callback(sub {


plicease:
👋🏻 any chance this can get a review, it is quite annoying on windows to see the color codes.

karenetheridge:
does loading Win32::Console::ANSI help?

plicease:
So if I set `PERL5OPT` to `-MWin32::Console::ANSI`, this is what I see:

image

On the up side the color codes are no longer printing out.  On the downside only one line got a different color, and that color makes it unreadable.  Not an entirely successful outcome I would still prefer to be able to turn the colors off altogether.

rjbs:
@plicease Try setting ANSI_COLORS_DISABLED=1

karenetheridge:
or NO_COLOR=1

plicease:
Those environment variables do seem to work on my non-wndows machines, and I would doublechck when I get back to my windows machine to make sure that it works there as well.  Which I suppose is a fine workaround.

I guess it perplexes me why `DZIL_COLOR=0` shouldn't turn off the colorization.  I would actually prefer to set the application specific environment variable in this case.

rjbs:
It should work, I agree, I just wanted to make sure nothing weirder was going on, and to give you a way to not suffer until I address the above.

plicease:
Ah thanks for the clarification!  Without an explicit comment acknowledgement, I misinterpreted these suggestions to imply that it should work the way it does right now.

Just to clarify, the reason I'd prefer DZIL_COLOR=0 to work rather than using the more general environment variables is because they could hide the behavior in other tools that I would rather see when I am testing them.

rjbs:
@plicease I apologize, I thought this was just an issue, and not a fix, so I put this off, then wrote my own fix, nearly identical to yours.  Since mine is already pushed, I am going to release it.  You have my thanks!

plicease:
> @plicease I apologize, I thought this was just an issue, and not a fix, so I put this off, then wrote my own fix, nearly identical to yours. Since mine is already pushed, I am going to release it. You have my thanks!

No problem.  I'm glad we got it sorted in the end.  I like your fix with `//` better anyway.  I will close #671, which was the issue where I initially raised this.

PR #498, opened 2015-10-06
* Proposed fix for missing 'use strict; use warnings;' 

Manwar:
Hi,

I propose fix to for missing 'use strict; use warnings;' line in the auto generated pod syntax test script.

Please review it.
Many Thanks.

Best Regards,
Mohammad S Anwar

PR #714, opened 2022-09-26
enable strict in generated author tests

mark-5:
Similar to #498, this enables strict and warning pragmas to 2 more author tests Dist-Zilla generates.


rjbs:
I amended your commit to put the two `use` statements on their own lines.  Thanks!

PR #715, opened 2022-11-08
Update dzil

rwp0:
package: enable GitHub syntax highlighting
L: http -> https

### DIFF
--- a/bin/dzil
+++ b/bin/dzil
@@ -1,6 +1,5 @@
 #!/usr/bin/perl
-package
-  dzil;
+package dzil;
 use Dist::Zilla::App;
 
 use Dist::Zilla::Pragmas;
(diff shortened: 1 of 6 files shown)


bin/dzil
```diff
@@ -1,6 +1,5 @@
 #!/usr/bin/perl
-package
-  dzil;
```
Grinnz, line 3: This is intentional and affects PAUSE's indexing of the package, so should not be changed.

rwp0: Could you please elaborate more on the PAUSE indexing? Thanks.

Grinnz: From the PAUSE operating model (https://pause.perl.org/pause/query?ACTION=pause_operating_model)

> In the source of the module, if there is a newline character between "package" and the package name, then PAUSE will take this as a hint to not consider the module for indexing (this was a hack put in place before the no_index directive was available, but given it is used by many distributions on CPAN, it will be respected for the foreseeable future).

rwp0: Thanks, I see now, perhaps it's time to switch to the `no_index` directive? :-)

rjbs:
I have applied the HTTP -> HTTPS update, then made more of the same sort, and dropped the bit that would affect PAUSE.

PR #716, opened 2022-12-03
Translate Dist::Zilla with Babble

Leont:
This is a proof of concept, not intended for merging into Dist::Zilla but to facilitate experimenting with Babble.

It currently produces a working Dist::Zilla, but not one that actually works on older perls because the translation isn't complete yet.

### DIFF
--- a/dist.ini
+++ b/dist.ini
@@ -87,3 +87,7 @@ Dist::Zilla::Plugin::NameFromDirectory = <= 0.03
 :version = 0.017
 conflicts_module = Moose::Conflicts
 conflicts_module = Module::Runtime::Conflicts
+
+[Babble]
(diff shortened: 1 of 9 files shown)


rjbs:
Maybe this proved a concept 3-4 years ago, but I'm not sure it's useful to keep this open.

karenetheridge:
I support the move from Try::Tiny to Feature::Compat::Try, but I don't generally make changes like that in my own distributions that aren't still under active development, so... `¯\_(ツ)_/¯`

Leont:
> Maybe this proved a concept 3-4 years ago, but I'm not sure it's useful to keep this open.

Fair, if Babble got any serious adoption it might have been interesting to pursue, but that does not seem to have materialized.

PR #717, opened 2022-12-09
Switch to use Feature::Compat::Try instead of Try::Tiny

Leont:
This makes it «use feature 'try'» on perl 5.36 and up, and use Syntax::Feature::Try on older versions.

### DIFF
--- a/lib/Dist/Zilla.pm
+++ b/lib/Dist/Zilla.pm
@@ -19,7 +19,7 @@ use Dist::Zilla::Path;
 use List::Util 1.33 qw(first none);
 use Software::License 0.104001; # ->program
 use String::RewritePrefix;
-use Try::Tiny;
+use Feature::Compat::Try;
 
 use Dist::Zilla::Prereqs;
 use Dist::Zilla::File::OnDisk;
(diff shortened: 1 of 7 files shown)


rjbs:
I don't know, because I'm not sure what the benefit of using this over Try::Tiny is, here.  What are we getting?

Grinnz:
Efficiency, ability to use `return` normally, avoidance of superglobals, similarity to the upcoming 'try' feature!

Grinnz:
Just make sure to use Syntax::Keyword::Try not the unrelated Syntax::Feature::Try 😄

karenetheridge:
> not the unrelated Syntax::Feature::Try

gah! I didn't realize that existed.. we should watch out for this.

Leont:
> Just make sure to use Syntax::Keyword::Try not the unrelated Syntax::Feature::Try smile

Yeah I keep confusing them, have to check the author to know which one I want

rjbs:
I'm getting pretty close to just saying "v5.36 minimum for Dist::Zilla", which may solve this problem by letting us move right to try.

PR #718, opened 2023-01-18
make dzil new options options take priority over config

haarg:
The options given to dzil new should take priority over options specified in the config file. This requires removing the default values given to App::Cmd so we can detect if they were specified or not.


rjbs:
Thanks!

PR #719, opened 2023-01-18
dzil add: follow %Mint config if no options given

haarg:
If a provider and profile are specified as options, fall back to using options from the %Mint stash before using the defaults. This will use the same config as the new command, but also allows that to be overridden in the dist's own dist.ini file.


rjbs:
Thanks!

PR #722, opened 2023-06-05
do not warn about 'uninitialized value' on non-symlink platforms

reneeb:
'dzil test' warns on some platforms about 'uninitialized value latest'. (see https://github.com/PerlServices/OPM-Maker-Command-depcheck/actions/runs/5146304937/jobs/9265154121#step:7:39)


rjbs:
Thanks, applied.

PR #728, opened 2025-03-11
avoid "wide character" warning when printing non-ascii sub names

### LINKED ISSUE #1031: wrong encoding on filehandle when printing an error
I got this in my Acme-Pi distribution while running release tests (before adding an appropriate pod directive to make the test pass):

```
#   Failed test 'Pod coverage on Acme::Pi'
#   at /Volumes/chambord/Users/ether/.perlbrew/libs/41.9@std/lib/perl5/Test/Pod/Coverage.pm line 133.
# Coverage for Acme::Pi is 0.0%, with 2 naked subroutines:
Wide character in print at /Volumes/chambord/Users/ether/perl5/perlbrew/perls/41.9/lib/5.41.9/Test2/Formatter/TAP.pm line 125.
# 	Ï
Wide character in print at /Volumes/chambord/Users/ether/perl5/perlbrew/perls/41.9/lib/5.41.9/Test2/Formatter/TAP.pm line 125.
# 	ð¿
# Looks like you failed 1 test of 1.
```

Adding `use open ':std', ':encoding(UTF-8)';` to the test (xt/author/pod-coverage.t) doesn't help.

This is with version 1.302209.

karenetheridge:
closes https://github.com/Test-More/test-more/issues/1031


Grinnz:
I don't think this is the best way to manipulate TAP output which relies on STDOUT, but I suppose it depends how Test::Pod::Coverage works.

karenetheridge:
I suppose we could do `use Test2::Plugin::UTF8;` instead, but that would force Test2 into the distribution -- albeit only for author tests, so maybe that's not a big deal.

Grinnz:
True, and it is a core module in recent Perls, but given the complication of injecting dependencies it's probably not worth it here. I would however suggest doing it this way so it is specifically targeted to STDOUT (assuming the relevant characters are in the TAP stream output):

```
{ use open ':std', OUT => ':encoding(UTF-8)' }
```

PR #731, opened 2025-04-13
autodie is lexically scoped, so must be used in each package

djerius:


djerius:
**Note! Behavior change:** switching `autodie` on in `Dist::Zilla::Tester::_Role` found a bug: the `rmdir` in `DEMOLISH` was not working as `$self->tempdir_root` was actually returning an error, as the directory was not empty.  This caused that directory to be left after the tests were completed.

The new code now uses `Dist::Zilla::Path->remove_tree` to clean up the path, so there's a behavior change: the `tmp` directory is no longer left.  Not sure if this is important.

haarg:
The subject isn't really correct.

If `autodie` was entirely lexically scoped, it would not be necessary to use it in each package. It's actually both lexically and package scoped. It exports subs to the calling package, but also sets a lexical flag to enable them.

djerius:
> The subject isn't really correct.
> 
> If `autodie` was entirely lexically scoped, it would not be necessary to use it in each package. It's actually both lexically and package scoped. It exports subs to the calling package, but also sets a lexical flag to enable them.

Ok, how about "packages which depend upon autodie must use it."?

rjbs:
I address this with an alternate take in https://github.com/rjbs/Dist-Zilla/pull/746

1. yes, import autodie where needed
2. do not turn failure to remove into fatal in tests: multiple tests share one tmpdir root, so they should be deleted when actually empty

remove_tree on tempdir_root is dangerous; caller could provide some dir to be used by n things

will ship fix in next release

djerius:
Thanks!

PR #732, opened 2025-04-17
use ustar format when using Archive::Tar::Wrapper

haarg:
Using the ustar format will avoid storing things like xattrs, which can throw warnings when being extracted on some systems.


rjbs:
Great, thanks!

PR #737, opened 2025-08-14
FAILING: Have DZ generate Makefile.PL with INSTALLDIRS

jkeenan:
**This pull request is submitted solely as a demonstration of a problem.  It is currently failing one test and is not mergeable.  Feedback requested.**

There exist several CPAN distributions which (a) are "dual-life" distributions shipped with the Perl 5 core underneath 'cpan/'; (b) in core, do not have their own Makefile.PL files; (c) have been determined to lack INSTALLDIRS key-value pairs in their upstream Makefile.PL files; and (d) lack an INSTALLDIRS kvp there because the Makefile.PL has been generated by Dist::Zilla::Plugin::Makemaker (or something that inherits therefrom).

Attempting to alleviate this problem going forwards, I herein tried to create a pull request in which I first added a test for the presence of INSTALLDIRS in a DZ-generated Makefile.PL (which at first failed, as expected), and then added code in Dist::Zilla::Plugin::Makemaker to insert such an INSTALLDIRS kvp.  However, I cannot get t/plugins/makemaker.t to PASS.  It repeatedly fails the first unit test therein.  I do not understand DZ enough to fix this failure.  See approx. line 62 of t/plugins/makemaker.t.

### DIFF
--- a/lib/Dist/Zilla/Plugin/MakeMaker.pm
+++ b/lib/Dist/Zilla/Plugin/MakeMaker.pm
@@ -260,6 +260,7 @@ sub write_makefile_args {
     VERSION   => $self->zilla->version,
     LICENSE   => $self->zilla->license->meta_yml_name,
     @exe_files ? ( EXE_FILES => [ sort @exe_files ] ) : (),
+    INSTALLDIRS => ($] < 5.011 ? 'perl' : 'site'),
 
     CONFIGURE_REQUIRES => $require_prereqs{configure},
     keys %{ $require_prereqs{build} } ? ( BUILD_REQUIRES => $require_prereqs{build} ) : (),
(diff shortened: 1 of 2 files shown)


karenetheridge:
> There exist several CPAN distributions which...

Which distributions are these?

Generally, we use the [\[DualLife\](https://metacpan.org/pod/Dist::Zilla::Plugin::DualLife) plugin to manage different install locations, but not all distributions need them because this construct is only relevant for perls before 5.011, and the distributions in question have a minimum perl that is later than that.

karenetheridge:
BTW, your branch passes just fine for me -- but this is not the correct change as we do not ever want to install non-dual-life modules into 'perl'. As stated before, we use the DualLife plugin to perform this selectively instead.

karenetheridge:
@jkeenan Could you please answer my question above? Which distributions have you seen to be showing issues?

PR #741, opened 2026-01-13
prevent Vim from detecting generated files as Prolog

haarg:
Vim tries to detect what file type .pl files are by checking their first line for a specific pattern. This pattern thinks lines ending with a period are Prolog files.

Remove the trailing period on the 'automatically generated' comments added to the various generated files to prevent this. All of the generated files are modified for consistency.


haarg:
I don't really care if this gets merged, as there are local workarounds. But since I tracked down what was causing this I figured I'd propose a possible solution.

rjbs:
It is absolutely unhinged that this works.

haarg:
The change that lead to this misdetection is https://github.com/vim/vim/commit/37853b7de31ef34153fe76aa2b740d517ed0e5d4

karenetheridge:
..which was PR https://github.com/vim/vim/pull/15253

karenetheridge:
[InstallGuide] is another plugin that might need an update. Strangely though I can't reproduce this issue locally (running vim 9.1 via macports, maybe too old?)

haarg:
I've worked around this for myself by setting
```vimrc
let g:filetype_pl = 'perl'
```
since I never deal with Prolog files.

rjbs:
I am not excited about making the text very slightly less grammatical, but whatever, I guess.  Like, I actually *like* Prolog, but…

haarg:
It would be nice if vim's detection mechanism could be improved, but I'm not familiar enough with prolog to be able to provide that kind of fix.

